PR #37, opened 2023-01-11
Adding on-chain quest fee to erc20 

GarrettJMU:


waynehoover (approved):
This looks good to me.

contracts/Erc20Quest.sol
```diff
@@ -15,28 +20,47 @@ contract Erc20Quest is Quest {
...
-        if (IERC20(rewardToken).balanceOf(address(this)) < totalAmount) revert TotalAmountExceedsBalance();
+        if (IERC20(rewardToken).balanceOf(address(this)) < (totalAmount + (totalAmount * questFee / 10_000))) revert TotalAmountExceedsBalance();
         super.start();
     }
 
+    function mintReceipt(uint amount_, string memory questId_, bytes32 hash_, bytes memory signature_) public {
+        if (hasStarted == false) revert NotStarted();
+        if (isPaused == true) revert QuestPaused();
+        if (block.timestamp < startTime) revert NotStarted();
+        if (block.timestamp > endTime) revert QuestEnded();
```
jonathandiep, line 54: Do we want the `QuestEnded` check on Erc1155Quest `claim` function?

contracts/QuestFactory.sol
```diff
@@ -23,6 +23,7 @@ contract QuestFactory is Initializable, OwnableUpgradeable, AccessControlUpgrade
 
     // storage vars. Insert new vars at the end to keep the storage layout the same.
     address public claimSignerAddress;
+    address public protocolFeeRecipient;
```
jonathandiep, line 26: I think we want to insert this at the end of the vars (check out comment on line 24)

@waynehoover Can you confirm?

GarrettJMU: Well we're going to deploy a new contract anyway, but I'm cool either way.

contracts/Erc20Quest.sol
GarrettJMU: Since we're the only ones running the 1155 Quests I'm not _as_ concerned about it.

contracts/QuestFactory.sol
jonathandiep: Hmm I think should be okay. We haven't set any quests on new contract so should be good

jonathandiep (approved):

PR #38, opened 2023-01-12
[1218] add access control

waynehoover:
Adds access control to the `createQuest` function, along with the `CREATE_QUEST_ROLE`.

Note: we will need to do a one time call of the `grantDefaultAdminAndCreateQuestRole` function to give the admin access role to the owner.

### DIFF
--- a/contracts/QuestFactory.sol
+++ b/contracts/QuestFactory.sol
@@ -7,15 +7,19 @@ import {Erc1155Quest} from './Erc1155Quest.sol';
 import {RabbitHoleReceipt} from './RabbitHoleReceipt.sol';
 import {OwnableUpgradeable} from '@openzeppelin/contracts-upgradeable/access/OwnableUpgradeable.sol';
 import '@openzeppelin/contracts-upgradeable/utils/cryptography/ECDSAUpgradeable.sol';
+import '@openzeppelin/contracts-upgradeable/access/AccessControlUpgradeable.sol';
 
-contract QuestFactory is Initializable, OwnableUpgradeable {
+contract QuestFactory is Initializable, OwnableUpgradeable, AccessControlUpgradeable {
     error QuestIdUsed();
     error OverMaxAllowedToMint();
     error AddressNotSigned();
     error AddressAlreadyMinted();
     error InvalidHash();
+    error InavlidRoleToCreateQuest();
 
     event QuestCreated(address indexed creator, address indexed contractAddress, string contractType);
+
+    bytes32 public constant CREATE_QUEST_ROLE = keccak256("CREATE_QUEST_ROLE");
 
     // storage vars. Insert new vars at the end to keep the storage layout the same.
     address public claimSignerAddress;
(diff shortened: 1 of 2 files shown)


test/QuestFactory.spec.ts
```diff
@@ -120,8 +120,19 @@ describe('QuestFactory', () => {
     })
 
     it('Should revert if trying to use existing quest id', async () => {
-      expect(
```
waynehoover, line 123: this test wasn't actually testing anything, so I fixed it here.

contracts/QuestFactory.sol
```diff
@@ -89,6 +97,19 @@ contract QuestFactory is Initializable, OwnableUpgradeable {
         return address(0);
     }
 
+    function grantCreateQuestRole(address account_) public {
```
waynehoover, line 100: no need to make this or `revokeCreateQuestRole` onlyAdmin because only the `DEFAULT_ADMIN_ROLE` can grant or revoke roles.

contracts/QuestFactory.sol
```diff
@@ -45,7 +51,9 @@ contract QuestFactory is Initializable, OwnableUpgradeable {
         string memory contractType_,
         string memory questId_,
         address receiptContractAddress_
-    ) public onlyOwner returns (address) {
+    ) public returns (address) {
+        if (!hasRole(CREATE_QUEST_ROLE, msg.sender)) revert InavlidRoleToCreateQuest();
```
jonathandiep, line 55: Any benefit for using this over using OZ's `onlyRole` modifier?

test/QuestFactory.spec.ts
jonathandiep: Ahhhhh thank you 😅🫡

contracts/QuestFactory.sol
```diff
@@ -7,16 +7,20 @@ import {Erc1155Quest} from './Erc1155Quest.sol';
...
     error QuestIdUsed();
     error OverMaxAllowedToMint();
     error AddressNotSigned();
     error AddressAlreadyMinted();
     error InvalidHash();
+    error InavlidRoleToCreateQuest();
 
     event QuestCreated(address indexed creator, address indexed contractAddress, string contractType);
 
+    bytes32 public constant CREATE_QUEST_ROLE = keccak256("CREATE_QUEST_ROLE");
```
jonathandiep, line 22: Doesn't inserting this var here mess up the storage layout?

contracts/QuestFactory.sol
```diff
@@ -31,6 +35,8 @@ contract QuestFactory is Initializable, OwnableUpgradeable {
 
     function initialize(address claimSignerAddress_, address rabbitholeReceiptContract_) public initializer {
         __Ownable_init();
+        __AccessControl_init();
+        grantDefaultAdminAndCreateQuestRole();
```
jonathandiep, line 39: Since we already deployed, I imagine that the `initialize` function is already called. Would we need to redeploy a new set of contracts?

contracts/QuestFactory.sol
waynehoover: no because it's a constant it's not a storage variable, the compiler treats it like a hardcoded variable.

contracts/QuestFactory.sol
waynehoover: we don't need to, we just need to call `grantDefaultAdminAndCreateQuestRole` after deployment.

contracts/QuestFactory.sol
waynehoover: no not really, the only reason was for the `InavlidRoleToCreateQuest` custom error, but I'm happy to move it to the modifier too, less code that way.

PR #43, opened 2023-01-19
Remove on-chain allowlist

jonathandiep:


contracts/Erc1155Quest.sol
```diff
@@ -11,21 +11,19 @@ contract Erc1155Quest is Quest, ERC1155Holder {
         uint256 endTime_,
         uint256 startTime_,
         uint256 totalAmount_,
-        string memory allowList_,
```
waynehoover, line 14: What happens if you just keep this here? I know it's not used but does it help compile?

jonathandiep: If I keep this, then I'll also need to keep the allowlist param in `Quest.sol` and in `QuestFactory.sol`

Which basically undos this PR lol

waynehoover: Why do you need to keep the param in the fx?

PR #44, opened 2023-01-19
Remove quantity in minting receipts, change protocolFee math, spec updates and some refactorings

waynehoover:
- Removed quantity to mint
- Fixed math and logic around protocolFee
- Added `struct Quest` to the QuestFactory contract
- lots of small refactors
  - changed `totalAmount_` to `totalParticipants_,`
  - `withdrawRemainingTokens` now takes an address

This requires a redeploy of the receipt and factory contracts.

### DIFF
--- a/contracts/Erc1155Quest.sol
+++ b/contracts/Erc1155Quest.sol
@@ -34,13 +34,6 @@ contract Erc1155Quest is Quest, ERC1155Holder {
         super.start();
     }
 
-    function claim() public override returns (uint256) {
-        if (IERC1155(rewardToken).balanceOf(address(this), rewardAmountInWeiOrTokenId) < totalParticipants)
-            revert AmountExceedsBalance();
-        super.claim();
-        return rewardAmountInWeiOrTokenId; // not used
-    }
-
     function _transferRewards(uint256 amount_) internal override {
         IERC1155(rewardToken).safeTransferFrom(address(this), msg.sender, rewardAmountInWeiOrTokenId, amount_, '0x00');
     }
--- a/contracts/Erc1155Quest.sol
+++ b/contracts/Erc1155Quest.sol
@@ -10,7 +10,7 @@ contract Erc1155Quest is Quest, ERC1155Holder {
         address rewardTokenAddress_,
         uint256 endTime_,
         uint256 startTime_,
-        uint256 totalAmount_,
+        uint256 totalParticipants_,
         string memory allowList_,
         uint256 rewardAmountInWeiOrTokenId_,
         string memory questId_,
@@ -20,7 +20,7 @@ contract Erc1155Quest is Quest, ERC1155Holder {
         rewardTokenAddress_,
         endTime_,
         startTime_,
-        totalAmount_,
+        totalParticipants_,
         allowList_,
         rewardAmountInWeiOrTokenId_,
         questId_,
@@ -29,15 +29,16 @@ contract Erc1155Quest is Quest, ERC1155Holder {
     {}
 
     function start() public override {
-        if (IERC1155(rewardToken).balanceOf(address(this), rewardAmountInWeiOrTokenId) < totalAmount)
+        if (IERC1155(rewardToken).balanceOf(address(this), rewardAmountInWeiOrTokenId) < totalParticipants)
             revert TotalAmountExceedsBalance();
         super.start();
     }
 
-    function claim() public override {
-        if (IERC1155(rewardToken).balanceOf(address(this), rewardAmountInWeiOrTokenId) < totalAmount)
+    function claim() public override returns (uint256) {
+        if (IERC1155(rewardToken).balanceOf(address(this), rewardAmountInWeiOrTokenId) < totalParticipants)
             revert AmountExceedsBalance();
         super.claim();
+        return rewardAmountInWeiOrTokenId; // not used
     }
 
     function _transferRewards(uint256 amount_) internal override {
@@ -48,11 +49,11 @@ contract Erc1155Quest is Quest, ERC1155Holder {
         return redeemableTokenCount_;
     }
 
-    function withdrawRemainingTokens() public override onlyOwner {
-        super.withdrawRemainingTokens();
+    function withdrawRemainingTokens(address to_) public override onlyOwner {
+        super.withdrawRemainingTokens(to_);
         IERC1155(rewardToken).safeTransferFrom(
             address(this),
-            msg.sender,
+            to_,
             rewardAmountInWeiOrTokenId,
             IERC1155(rewardToken).balanceOf(address(this), rewardAmountInWeiOrTokenId),
             '0x00'
--- a/contracts/Erc20Quest.sol
+++ b/contracts/Erc20Quest.sol
@@ -53,13 +53,6 @@ contract Erc20Quest is Quest {
         super.start();
     }
 
-    function claim() public override returns (uint256) {
-        uint redeemableTokenCount = super.claim();
-        reedemedTokens += redeemableTokenCount;
-
-        return redeemableTokenCount; // not used
-    }
-
     function _transferRewards(uint256 amount_) internal override {
         IERC20(rewardToken).safeTransfer(msg.sender, amount_);
     }
--- a/contracts/Erc20Quest.sol
+++ b/contracts/Erc20Quest.sol
@@ -77,7 +77,7 @@ contract Erc20Quest is Quest {
     }
 
     function receiptRedeemers() public view returns (uint256) {
-        return questFactoryContract.numberMintedForQuestId(questId);
+        return questFactoryContract.getNumberMinted(questId);
     }
 
     function protocolFee() public view returns (uint256) {
--- a/contracts/Erc20Quest.sol
+++ b/contracts/Erc20Quest.sol
@@ -16,7 +16,7 @@ contract Erc20Quest is Quest {
         address rewardTokenAddress_,
         uint256 endTime_,
         uint256 startTime_,
-        uint256 totalAmount_,
+        uint256 totalParticipants_,
         string memory allowList_,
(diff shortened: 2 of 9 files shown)


test/Erc20Quest.spec.ts
```diff
@@ -343,59 +355,67 @@ describe('Erc20Quest', async () => {
...
+
+      await deployedQuestContract.connect(firstAddress).claim()
+      expect(await deployedSampleErc20Contract.balanceOf(firstAddress.address)).to.equal(100)
+      expect(beginningContractBalance).to.equal(totalRewardsPlusFee)
+
+      await ethers.provider.send('evm_increaseTime', [100001])
+      await deployedQuestContract.withdrawFee()
+
+      expect(await deployedSampleErc20Contract.balanceOf(deployedQuestContract.address)).to.equal(1100)
+      expect(await deployedSampleErc20Contract.balanceOf(protocolFeeAddress)).to.equal(0) // I don't think this math is correct
```
waynehoover, line 415: we should have some protocol fees here, I think the math is incorrect.

contracts/Erc20Quest.sol
```diff
@@ -29,38 +29,41 @@ contract Erc20Quest is Quest {
...
-        super.start();
+    function questFeePercentage() public view returns (uint256) {
+        return questFee / 10_000;
+    }
+
+    function maxTotalRewards() public view returns (uint256) {
+        return totalParticipants * rewardAmountInWeiOrTokenId;
     }
 
-    function mintReceipt(uint amount_, string memory questId_, bytes32 hash_, bytes memory signature_) public {
```
waynehoover, line 50: @jonathandiep I removed this function here, so it continues to live on the factory contract.

contracts/QuestFactory.sol
```diff
@@ -150,6 +151,7 @@ contract QuestFactory is Initializable, OwnableUpgradeable, AccessControlUpgrade
 
         amountMintedForQuestId[questId_] += amount_;
         addressMintedForQuestId[questId_][msg.sender] = true;
+        numberMintedForQuestId[questId_]++;
```
jonathandiep, line 154: Isn't this a duplicate of `amountMintedForQuestId`?

contracts/Erc20Quest.sol
```diff
@@ -29,38 +29,41 @@ contract Erc20Quest is Quest {
...
-        totalAmount_,
+        totalParticipants_,
         allowList_,
         rewardAmountInWeiOrTokenId_,
         questId_,
         receiptContractAddress_
     ) {
         questFee = questFee_;
-        totalRedeemers = rewardAmountInWeiOrTokenId / totalAmount;
+        totalRedeemers = totalParticipants / rewardAmountInWeiOrTokenId_;
```
jonathandiep, line 39: I don't think we need `totalRedeemers` anymore!

test/Erc20Quest.spec.ts
```diff
@@ -307,7 +320,7 @@ describe('Erc20Quest', async () => {
       await ethers.provider.send('evm_increaseTime', [86400])
 
       const startingBalance = await deployedSampleErc20Contract.balanceOf(owner.address)
-      expect(startingBalance.toString()).to.equal('0')
+      expect(startingBalance).to.equal(0)
```
jonathandiep, line 323: Don't the values from the contract queries come back as BigNumber?

contracts/Erc20Quest.sol
```diff
@@ -71,15 +70,24 @@ contract Erc20Quest is Quest {
...
 
-        uint256 nonClaimableTokens = (totalRedeemers - receiptRedeemers) * rewardAmountInWeiOrTokenId;
-        IERC20(rewardToken).safeTransfer(msg.sender, nonClaimableTokens);
+    function protocolFee() public view returns (uint256) {
+        return receiptRedeemers() * rewardAmountInWeiOrTokenId * questFee / 10_000;
     }
 
-    function withdrawFee() public {
-        IERC20(rewardToken).safeTransfer(protocolFeeRecipient, (receiptRedeemers * rewardAmountInWeiOrTokenId * questFee / 10_000));
+    function withdrawFee() public onlyStarted {
```
jonathandiep, line 89: I don't think it matters much (since `protocolFee()` is amount we can take), but should we check that quest is finished?

Or should we keep as is and withdraw as often as we please?

test/Erc20Quest.spec.ts
```diff
@@ -18,11 +19,13 @@ describe('Erc20Quest', async () => {
...
   let expiryDate: number, startDate: number
   const mockAddress = '0x0000000000000000000000000000000000000000'
+  const mnemonic = 'announce room limb pattern dry unit scale effort smooth jazz weasel alcohol'
   const questId = 'asdf'
   const allowList = 'ipfs://someCidToAnArrayOfAddresses'
-  const totalRewards = 1000
-  const totalRewardsPlusFee = 1200
-  const rewardAmount = 10
+  const totalParticipants = 300
+  const totalRewardsPlusFee = 120000
```
jonathandiep, line 26: Shouldn't `totalRewardsPlusFee` be

```
totalParticipants * rewardAmount + (totalParticipants * rewardAmount * questFee / 10_000)
```
which in this test case would be
```
300 * 1000 + (300 * 1000  * 2000 / 10000)
```

test/Erc20Quest.spec.ts
jonathandiep: Awesome, ran the tests and logged these and this seems to work for some reason lol. If no need to convert lets keep it simple!

test/Erc20Quest.spec.ts
```diff
@@ -316,15 +329,15 @@ describe('Erc20Quest', async () => {
 
       await deployedQuestContract.claim()
       const endingBalance = await deployedSampleErc20Contract.balanceOf(owner.address)
-      expect(endingBalance.toString()).to.equal('10')
+      expect(endingBalance).to.equal(1000)
```
jonathandiep, line 332: Where 1000 is used I think it makes more sense to use `rewardAmount`

test/Erc20Quest.spec.ts
waynehoover: Yes it should, thanks for pointing this out. I updated the tests to use this math.

contracts/Erc20Quest.sol
waynehoover: @GarrettJMU what do you think? I think it's fine to withdraw the protocolFee whenever.

contracts/QuestFactory.sol
waynehoover: Kind of but not really, but it is a good question on if we should use `amountMintedForQuestId` instead of `numberMintedForQuestId`.

contracts/Erc1155Quest.sol
```diff
@@ -20,7 +20,7 @@ contract Erc1155Quest is Quest, ERC1155Holder {
         rewardTokenAddress_,
         endTime_,
         startTime_,
-        totalAmount_,
+        totalParticipants_,
```
GarrettJMU, line 23: so much better 😭

GarrettJMU (approved):
A few small comments

jonathandiep (approved):
Oooo I love the structs!

contracts/Erc20Quest.sol
```diff
@@ -71,15 +68,24 @@ contract Erc20Quest is Quest {
...
+    }
+
+    function receiptRedeemers() public view returns (uint256) {
+        return questFactoryContract.getNumberMinted(questId);
+    }
 
-        uint256 nonClaimableTokens = (totalRedeemers - receiptRedeemers) * rewardAmountInWeiOrTokenId;
-        IERC20(rewardToken).safeTransfer(msg.sender, nonClaimableTokens);
+    function protocolFee() public view returns (uint256) {
+        return receiptRedeemers() * rewardAmountInWeiOrTokenId * questFee / 10_000;
```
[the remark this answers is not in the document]
waynehoover: We can't because that will return a unit, which would be less than zero, so it would return zero.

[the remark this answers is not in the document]
GarrettJMU: ooh that's right 👍🏼

contracts/Erc1155Quest.sol
```diff
@@ -29,15 +29,16 @@ contract Erc1155Quest is Quest, ERC1155Holder {
...
     function start() public override {
-        if (IERC1155(rewardToken).balanceOf(address(this), rewardAmountInWeiOrTokenId) < totalAmount)
+        if (IERC1155(rewardToken).balanceOf(address(this), rewardAmountInWeiOrTokenId) < totalParticipants)
             revert TotalAmountExceedsBalance();
         super.start();
     }
 
-    function claim() public override {
-        if (IERC1155(rewardToken).balanceOf(address(this), rewardAmountInWeiOrTokenId) < totalAmount)
+    function claim() public override returns (uint256) {
```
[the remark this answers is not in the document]
waynehoover: good call, we can get rid of it.

contracts/Erc20Quest.sol
```diff
@@ -29,38 +28,36 @@ contract Erc20Quest is Quest {
...
+        if (IERC20(rewardToken).balanceOf(address(this)) < maxTotalRewards() + maxProtocolReward()) revert TotalAmountExceedsBalance();
+        super.start();
     }
 
-    function claim() public override {
-        if (IERC20(rewardToken).balanceOf(address(this)) < (rewardAmountInWeiOrTokenId * questFee / 10_000)) revert AmountExceedsBalance();
-        super.claim();
-        rewardRedeemers++;
+    function claim() public override returns (uint256) {
+        uint redeemableTokenCount = super.claim();
```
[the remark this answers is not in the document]
waynehoover: yep.

PR #57, opened 2023-01-21
Slither cleanup

GarrettJMU:
- Immutable variables
- Cleanup usage of params


jonathandiep (approved):
How did you get allowlist removal to work?

PR #60, opened 2023-01-22
Add Erc20Quest natspec comments

jonathandiep:


contracts/Erc20Quest.sol
```diff
@@ -22,41 +25,58 @@ contract Erc20Quest is Quest {
...
 
+    /// @dev Function that gets the maximum amount of rewards that can be claimed by the protocol or the quest deployer
+    /// @return The maximum amount of rewards that can be claimed by the protocol or the quest deployer
     function maxProtocolReward() public view returns (uint256) {
-        return maxTotalRewards() * questFee / 10_000;
+        return (maxTotalRewards() * questFee) / 10_000;
     }
 
+    /// @notice Starts the quest
+    /// @dev Requires that the balance of the rewards in the contract is greater than or equal to the maximum amount of rewards that can be claimed by all users and the protocol
```
GarrettJMU, line 56: we should also probably note here that it marks it as ready to start. It still needs to be past the start date

contracts/Erc20Quest.sol
```diff
@@ -22,41 +25,58 @@ contract Erc20Quest is Quest {
...
     }
 
+    /// @notice Internal function that transfers the rewards to the msg.sender
+    /// @param amount_ The amount of rewards to transfer
     function _transferRewards(uint256 amount_) internal override {
         IERC20(rewardToken).safeTransfer(msg.sender, amount_);
     }
 
+    /// @notice Internal function that calculates the reward amount
+    /// @dev It is possible for users to have multiple receipts
```
GarrettJMU, line 70: we should add in here if they buy others on secondary

contracts/Erc20Quest.sol
```diff
@@ -22,41 +25,58 @@ contract Erc20Quest is Quest {
...
+    /// @param redeemableTokenCount_ The amount of tokens that can be redeemed
+    /// @return The total amount of rewards that can be claimed by a user
     function _calculateRewards(uint256 redeemableTokenCount_) internal view override returns (uint256) {
         return redeemableTokenCount_ * rewardAmountInWeiOrTokenId;
     }
 
+    /// @notice Function that allows the owner to withdraw the remaining tokens in the contract
+    /// @dev Every receipt minted should still be able to claim rewards (and cannot be withdrawn)
```
GarrettJMU, line 78: we should also note this is after quest end time

contracts/Erc20Quest.sol
```diff
@@ -65,17 +85,20 @@ contract Erc20Quest is Quest {
...
     }
 
+    /// @notice Function that calculates the protocol fee
     function protocolFee() public view returns (uint256) {
-        return receiptRedeemers() * rewardAmountInWeiOrTokenId * questFee / 10_000;
+        return (receiptRedeemers() * rewardAmountInWeiOrTokenId * questFee) / 10_000;
     }
 
+    /// @notice Sends the protocol fee to the protocolFeeRecipient
+    /// @dev Only callable when the quest is active and started
```
GarrettJMU, line 100: honestly we should move this to when it's ended. Doesn't make sense to withdraw before

jonathandiep: Alright, I've changed the modifier so that it uses `onlyAdminWithdrawAfterEnd` rather than `onlyQuestActive`

contracts/Erc20Quest.sol
jonathandiep: Technically, we can call `start` before the start date (there's no check for it). Is that something we need to implement? I think we should also allow quest admins to start the quest whenever

GarrettJMU (approved):

PR #61, opened 2023-01-22
Add Quest natspec comments

jonathandiep:


contracts/Quest.sol
```diff
@@ -100,17 +116,25 @@ contract Quest is Ownable, IQuest {
         emit Claimed(msg.sender, totalRedeemableRewards);
```
jonathandiep, line 116: These lines are marked as unreachable (cause `_calculateRewards()` reverts here lol)

Don't think it'll be an issue on child contracts, but want to make sure

CleanShot 2023-01-22 at 02 33 06@2x

GarrettJMU: Yeah we should consider adding those functions to the interface instead tbh. I just had reverts there bc that's a common web2 pattern to force implementation

GarrettJMU (approved):

PR #69, opened 2023-01-23
add events to emit on ticket and receipt contracts

waynehoover:


contracts/interfaces/IQuestFactory.sol
```diff
@@ -13,5 +13,6 @@ interface IQuestFactory {
     error AddressZeroNotAllowed();
     error QuestFeeTooHigh();
 
-    event QuestCreated(address indexed creator, address indexed contractAddress, string contractType);
+
+    event QuestCreated(address indexed creator, address indexed contractAddress, string indexed questId, string contractType, address rewardTokenAddress, uint256 endTime, uint256 startTime, uint256 totalParticipants, uint256 rewardAmountOrTokenId);
```
waynehoover, line 17: is it more important to index questId or contractType? I would think questId.

GarrettJMU (approved):

PR #74, opened 2023-01-24
add more traits to receipts

waynehoover:
- adds trait metadata to the reciept renderer

Todo: tests

### DIFF
--- a/contracts/Quest.sol
+++ b/contracts/Quest.sol
@@ -136,6 +136,14 @@ contract Quest is Ownable, IQuest {
         return claimedList[tokenId_] == true;
     }
 
+    function getRewardAmount() public view returns (uint256) {
+        return rewardAmountInWeiOrTokenId;
+    }
+
+    function getRewardToken() public view returns (address) {
+        return rewardToken;
+    }
+
     /// @notice Allows the owner of the Quest to withdraw any remaining rewards after the Quest has ended
     function withdrawRemainingTokens(address to_) public virtual onlyOwner onlyAdminWithdrawAfterEnd {}
 }
--- a/contracts/Quest.sol
+++ b/contracts/Quest.sol
@@ -136,10 +136,12 @@ contract Quest is Ownable, IQuest {
         return claimedList[tokenId_] == true;
     }
 
+    /// @dev Returns the reward amount
     function getRewardAmount() public view returns (uint256) {
         return rewardAmountInWeiOrTokenId;
     }
 
+    /// @dev Returns the reward token address
     function getRewardToken() public view returns (address) {
         return rewardToken;
     }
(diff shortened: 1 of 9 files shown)


contracts/Quest.sol
```diff
@@ -136,6 +136,16 @@ contract Quest is Ownable, IQuest {
...
     }
 
+    /// @dev Returns the reward amount
+    function getRewardAmount() public view returns (uint256) {
+        return rewardAmountInWeiOrTokenId;
+    }
+
+    /// @dev Returns the reward token address
+    function getRewardToken() public view returns (address) {
+        return rewardToken;
```
GarrettJMU, line 146: same note here

contracts/RabbitHoleReceipt.sol
```diff
@@ -69,6 +72,12 @@ contract RabbitHoleReceipt is
         royaltyRecipient = royaltyRecipient_;
     }
 
+    /// @dev set the quest factory contract
+    /// @param questFactory_ the address of the quest factory contract
+    function setQuestFactory(address questFactory_) public onlyOwner {
+        QuestFactoryContract = IQuestFactory(questFactory_);
```
GarrettJMU, line 78: Is it more common to use Interface here or the actual contract?

contracts/Quest.sol
```diff
@@ -136,6 +136,16 @@ contract Quest is Ownable, IQuest {
         return claimedList[tokenId_] == true;
     }
 
+    /// @dev Returns the reward amount
+    function getRewardAmount() public view returns (uint256) {
+        return rewardAmountInWeiOrTokenId;
```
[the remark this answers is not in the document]
waynehoover: yes, but I'm not sure how to access it via the interface only. If I include the contract we get a circular dependency and it won't compile.

contracts/QuestFactory.sol
```diff
@@ -170,6 +170,16 @@ contract QuestFactory is Initializable, OwnableUpgradeable, AccessControlUpgrade
...
     }
 
+    /// @dev return data in the quest struct for a questId
+    /// @param questId_ The id of the quest
+    function questInfo(string memory questId_) external view returns (address, uint, uint) {
+        return (
+            quests[questId_].questAddress,
+            quests[questId_].totalParticipants,
+            quests[questId_].numberMinted
+        );
```
jonathandiep, line 180: I like this! But haven't really worked with tuples on the ethers.js side so interesting to see how the data will return (maybe a JS object?)

contracts/RabbitHoleReceipt.sol
waynehoover: the actual contract would have been better, but it creates a circular dependency and I couldn't.

contracts/QuestFactory.sol
jonathandiep: Nvm I see this mostly for RabbitHoleReceipt.sol

jonathandiep (approved):
lgtm

PR #77, opened 2023-01-24
Use ROLE_ACCESS role

jonathandiep:


contracts/QuestFactory.sol
```diff
@@ -139,19 +140,31 @@ contract QuestFactory is Initializable, OwnableUpgradeable, AccessControlUpgrade
...
         } else {
             _revokeRole(CREATE_QUEST_ROLE, account_);
         }
     }
 
-    /// @dev grant the default admin role and the create quest role to the owner
+    /// @dev grant the role access role to an account
+    /// @param account_ The account to grant or revoke the role access role to
+    /// @param canAccessRole_ Boolean to grant or revoke the role access role. True grants access
+    function changeRoleAccessRole(address account_, bool canAccessRole_) public onlyRole(ROLE_ACCESS) {
```
GarrettJMU, line 154:
```suggestion
    function changeRoleAccessRole(address account_, bool canAccessRole_) public onlyOwner {
```

jonathandiep:
Re-doing this. Will get rid of onlyOwner and use roles instead

PR #78, opened 2023-01-26
uint default value is 0

SharifElfouly:
uint default value is 0, so there is no need in setting it explicitly


waynehoover:
Thanks for this PR, this change was recently merged via another PR.

PR #81, opened 2023-01-31
Latest Deployments

jonathandiep:
Also added a view function

### DIFF
--- a/contracts/QuestFactory.sol
+++ b/contracts/QuestFactory.sol
@@ -136,17 +136,6 @@ contract QuestFactory is Initializable, OwnableUpgradeable, AccessControlUpgrade
         revert QuestTypeInvalid();
     }
 
-    /// @dev grant the create quest role to an account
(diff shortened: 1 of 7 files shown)


waynehoover (approved):

contracts/QuestFactory.sol
```diff
@@ -136,17 +136,6 @@ contract QuestFactory is Initializable, OwnableUpgradeable, AccessControlUpgrade
         revert QuestTypeInvalid();
     }
 
-    /// @dev grant the create quest role to an account
```
waynehoover, line 139: we don't need this anymore?

jonathandiep: AccessControl has native `grantRole` and `revokeRole`!

https://docs.openzeppelin.com/contracts/4.x/api/access#AccessControl-grantRole-bytes32-address-

GarrettJMU: ➕ 1

PR #82, opened 2023-02-01
Add Final SVG to renderer contract

waynehoover:
This adds the final SVG to the ReceiptRenderer contract.


hardhat.config.ts
```diff
@@ -43,6 +43,10 @@ const config: HardhatUserConfig = {
   networks: {
     hardhat: {
       chainId: 1337,
+      forking: {
```
waynehoover, line 46: moved to a forking test suite because we need to test on real erc20 addresses to test the receipt renderer `humanRewardAmountAndSymbol` function.

jonathandiep: Don't we need to run a different command when we wanna test fork mainnet? Or does running `yarn test` automatically apply the mainnet fork?

test/Erc1155Quest.spec.ts
```diff
@@ -41,8 +43,8 @@ describe('Erc1155Quest', () => {
     thirdAddress = local_thirdAddress
     fourthAddress = local_fourthAddress
 
-    expiryDate = Math.floor(Date.now() / 1000) + 10000
-    startDate = Math.floor(Date.now() / 1000) + 1000
+    expiryDate = latestTime + 1000
+    startDate = latestTime + 10
```
jonathandiep, line 47: Much more readable! NIce

jonathandiep: Oh what the real benefit of this is that you don't need to reverse the time. NICE

jonathandiep (approved):
LGTM!

I think the only thing left is to use the `questIdCount`? Still need to try fitting that in?

GarrettJMU (approved):

hardhat.config.ts
waynehoover: `yarn test` will now always apply the mainnet fork, yes.

waynehoover:
> LGTM!
> 
> I think the only thing left is to use the `questIdCount`? Still need to try fitting that in?

Yes, we need to decrease the contract size somehow to get the mapping to fit in first.

PR #83, opened 2023-02-03
New SVGs without questid in svg

waynehoover:
updates the renderer for new svgs without the questid.


jonathandiep (approved):
Awesome! I think we can remove the `questIdCount` as well

jonathandiep:
Also maybe another small afterthought about tests: should we test pre-claim render? We might also want to throw in the pre-claimed svg (the picture) into the repo as well

PR #84, opened 2023-02-06
deploy quests with minimal proxy

waynehoover:
- use OZ clones to clone either an 1155 or 721 Quest from addresses set or supplied on init.
- replace constructors with an `initialize` fx using the `initializer` modifier.
- replace the constructor from the base Quest contract with `questInit` using the `onlyInitializing` modifier.
- modify tests to deploy 1155 and 721 quests then pass addresses into questFactory
- modify questFactory deploy script


GarrettJMU:
Let's init implementation on deploy

GarrettJMU:
Let's also add a deploy1155 and deploy721 scripts

PR #88, opened 2023-02-08
[DEV-1606] - Enforce Quest start/end times for mint

jonathandiep:


contracts/QuestFactory.sol
```diff
@@ -231,6 +232,9 @@ contract QuestFactory is Initializable, OwnableUpgradeable, AccessControlUpgrade
     function mintReceipt(string memory questId_, bytes32 hash_, bytes memory signature_) public {
         if (quests[questId_].numberMinted + 1 > quests[questId_].totalParticipants) revert OverMaxAllowedToMint();
         if (quests[questId_].addressMinted[msg.sender] == true) revert AddressAlreadyMinted();
+        if (!QuestContract(quests[questId_].questAddress).hasStarted()) revert QuestNotStarted();
+        if (block.timestamp < QuestContract(quests[questId_].questAddress).startTime()) revert QuestNotStarted();
+        if (block.timestamp > QuestContract(quests[questId_].questAddress).endTime()) revert QuestEnded();
```
GarrettJMU, line 237: should we just create a struct on the quest contract that has these variables and return the struct here so we don't have to make 3 calls?

contracts/QuestFactory.sol
```diff
@@ -2,8 +2,9 @@
 pragma solidity ^0.8.15;
 
 import '@openzeppelin/contracts-upgradeable/proxy/utils/Initializable.sol';
-import {Erc20Quest} from './Erc20Quest.sol';
 import {IQuestFactory} from './interfaces/IQuestFactory.sol';
+import {Quest as QuestContract} from './Quest.sol';
```
waynehoover, line 6: why import the base contract? You can call all these functions from the `Erc20Quest` contract right?

jonathandiep: I think I was just worried for 1155 quests lol (even though they are quite similar)

I can change to `Erc20Quest`

contracts/QuestFactory.sol
jonathandiep: Just following up here, will keep as is as it doesn't really save on gas to make these calls

PR #94, opened 2023-02-10
Modify event params

jonathandiep:
Adds more details to the events that we care about. Removed indexes on `questId` as it turns into a hash and it's difficult to parse the actual value

### DIFF
--- a/contracts/QuestFactory.sol
+++ b/contracts/QuestFactory.sol
@@ -249,7 +249,7 @@ contract QuestFactory is Initializable, OwnableUpgradeable, AccessControlUpgrade
 
         quests[questId_].addressMinted[msg.sender] = true;
         quests[questId_].numberMinted++;
-        emit ReceiptMinted(msg.sender, questId_);
         rabbitHoleReceiptContract.mint(msg.sender, questId_);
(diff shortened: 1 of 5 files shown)


contracts/QuestFactory.sol
```diff
@@ -249,7 +249,7 @@ contract QuestFactory is Initializable, OwnableUpgradeable, AccessControlUpgrade
 
         quests[questId_].addressMinted[msg.sender] = true;
         quests[questId_].numberMinted++;
-        emit ReceiptMinted(msg.sender, questId_);
         rabbitHoleReceiptContract.mint(msg.sender, questId_);
+        emit ReceiptMinted(msg.sender, quests[questId_].questAddress, rabbitHoleReceiptContract.getTokenId(), questId_);
```
jonathandiep, line 253: Seeing the reentrancy attack from Slither here. I'll move the event before `mint()`, but will need to get token count and increase by 1 to get the correct tokenId

waynehoover: I think it's fine to have the event emit after contract call, because we are not setting any storage in the emit.

waynehoover (approved):

GarrettJMU (approved):

PR #95, opened 2023-02-10
Gas optimizations

GarrettJMU:
Quest.sol
- Moving from x += y to x = x+y
- Moving to unchecked for bound


contracts/Quest.sol
```diff
@@ -105,18 +106,19 @@ contract Quest is OwnableUpgradeable, IQuest {
         if (tokens.length == 0) revert NoTokensToClaim();
 
         uint256 redeemableTokenCount = 0;
-        for (uint i = 0; i < tokens.length; i++) {
+        for (uint i = 0; i < tokens.length;) {
             if (!isClaimed(tokens[i])) {
-                redeemableTokenCount++;
+            unchecked{redeemableTokenCount++;}
```
GarrettJMU, line 111: unsure about this one. I guess we should prob be safe for overflow here

contracts/Quest.sol
```diff
@@ -105,18 +106,19 @@ contract Quest is OwnableUpgradeable, IQuest {
...
             }
+        unchecked{i++;}
```
GarrettJMU, line 113: these are bounded, so should be fine

contracts/QuestFactory.sol
```diff
@@ -76,7 +76,7 @@ contract QuestFactory is Initializable, OwnableUpgradeable, AccessControlUpgrade
         uint256 rewardAmountOrTokenId_,
         string memory contractType_,
         string memory questId_
-    ) public onlyRole(CREATE_QUEST_ROLE) returns (address) {
+    ) external payable onlyRole(CREATE_QUEST_ROLE) returns (address) {
```
jonathandiep, line 79: Why introduce `payable`? Don't think anyone ever needs to send ETH to create a quest

GarrettJMU: If a function modifier such as onlyOwner is used, the function will revert if a normal user tries to pay the function. Marking the function as payable will lower the gas cost for legitimate callers because the compiler will not include checks for whether a payment was provided. 

The extra opcodes avoided are
CALLVALUE(2),DUP1(3),ISZERO(3),PUSH2(3),JUMPI(10),PUSH1(3),DUP1(3),REVERT(0),JUMPDEST(1),POP(2), which costs an average of about 21 gas per call to the function, in addition to the extra deployment cost.

contracts/Quest.sol
```diff
@@ -45,7 +45,6 @@ contract Quest is OwnableUpgradeable, IQuest {
         rewardAmountInWeiOrTokenId = rewardAmountInWeiOrTokenId_;
         questId = questId_;
         rabbitHoleReceiptContract = RabbitHoleReceipt(receiptContractAddress_);
-        redeemedTokens = 0;
```
GarrettJMU, line 48: init to 0 by default

contracts/Quest.sol
waynehoover: Yes, this is fine to me.

contracts/RabbitHoleReceipt.sol
```diff
@@ -161,7 +162,7 @@ contract RabbitHoleReceipt is
         require(_exists(tokenId_), 'ERC721URIStorage: URI query for nonexistent token');
         require(QuestFactoryContract != IQuestFactory(address(0)), 'QuestFactory not set');
 
-        string memory questId = questIdForTokenId[tokenId_];
+        string storage questId = questIdForTokenId[tokenId_];
```
waynehoover, line 165: `storage` is less gas than `memory` whys that?

waynehoover (approved):
All looks good to me.

contracts/QuestFactory.sol
```diff
@@ -76,10 +79,10 @@ contract QuestFactory is Initializable, OwnableUpgradeable, AccessControlUpgrade
         uint256 rewardAmountOrTokenId_,
         string memory contractType_,
         string memory questId_
-    ) public onlyRole(CREATE_QUEST_ROLE) returns (address) {
-        if (quests[questId_].questAddress != address(0)) revert QuestIdUsed();
-
-        if (keccak256(abi.encodePacked(contractType_)) == keccak256(abi.encodePacked('erc20'))) {
+    ) external payable onlyRole(CREATE_QUEST_ROLE) returns (address) {
+        currentQuest = quests[questId_];
```
jonathandiep, line 83: Do we not have to define a type/struct for `currentQuest`?

contracts/QuestFactory.sol
```diff
@@ -239,16 +243,17 @@ contract QuestFactory is Initializable, OwnableUpgradeable, AccessControlUpgrade
...
+        if (!QuestContract(currentQuest.questAddress).hasStarted()) revert QuestNotStarted();
+        if (block.timestamp < QuestContract(currentQuest.questAddress).startTime()) revert QuestNotStarted();
+        if (block.timestamp > QuestContract(currentQuest.questAddress).endTime()) revert QuestEnded();
         if (keccak256(abi.encodePacked(msg.sender, questId_)) != hash_) revert InvalidHash();
         if (recoverSigner(hash_, signature_) != claimSignerAddress) revert AddressNotSigned();
 
-        quests[questId_].addressMinted[msg.sender] = true;
-        quests[questId_].numberMinted++;
+        currentQuest.addressMinted[msg.sender] = true;
+        currentQuest.numberMinted++;
```
jonathandiep, line 256: I think we can save a little bit of gas on this: https://cygaar.substack.com/p/why-i-is-cheaper-than-i-in-solidity

Just need to consider how we use it (which I don't think is an issue here): https://ethereum.stackexchange.com/questions/133161/why-does-i-cost-less-gas-than-i/133164#133164

contracts/RabbitHoleReceipt.sol
jonathandiep, line 165: I'll need to read up on storage vs memory. I thought memory would make sense here cause `questId` is a temp variable that only exists inside this function and doesn't need to be stored anywhere

contracts/ReceiptRenderer.sol
```diff
@@ -46,11 +46,13 @@ contract ReceiptRenderer {
         uint rewardAmount_,
         address rewardAddress_
     ) public view virtual returns (bytes memory) {
+        tokenIdString = tokenId_.toString();
```
jonathandiep, line 49: Same q as earlier https://github.com/rabbitholegg/quest-protocol/pull/95/files#r1103266935

contracts/RabbitHoleReceipt.sol
GarrettJMU: Actually I don't think this makes sense tbh. Undoing

contracts/QuestFactory.sol
```diff
@@ -77,9 +80,11 @@ contract QuestFactory is Initializable, OwnableUpgradeable, AccessControlUpgrade
         string memory contractType_,
         string memory questId_
     ) external onlyRole(CREATE_QUEST_ROLE) returns (address) {
-        if (quests[questId_].questAddress != address(0)) revert QuestIdUsed();
+        Quest storage currentQuest = quests[questId_];
+        if (currentQuest.questAddress != address(0)) revert QuestIdUsed();
+        ++currentQuest.numberMinted;
```
jonathandiep, line 85: This should not be added here! Causes `numberMinted` to start from 1 instead of 0

jonathandiep (changes requested):

package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "quest_merkle_distributor",
+  "name": "quest-protocol",
```
jonathandiep, line 2: Haha nice!

Can we also change line 28?

jonathandiep (approved):

PR #96, opened 2023-02-10
move most public functions to external

waynehoover:
Most public functions have been made external to save gas. Some functions in `QuestFactory` gave a cryptic error(1) when converting to external so those have stayed public for now.

1: `VM Exception while processing transaction: reverted with reason string 'Address: low-level delegate call failed'`


contracts/Quest.sol
```diff
@@ -136,26 +136,23 @@ contract Quest is OwnableUpgradeable, IQuest {
...
     }
 
     /// @dev Returns the reward token address
-    function getRewardToken() public view returns (address) {
+    function getRewardToken() external view returns (address) {
         return rewardToken;
     }
 
-    /// @notice Allows the owner of the Quest to withdraw any remaining rewards after the Quest has ended
-    function withdrawRemainingTokens() public virtual onlyWithdrawAfterEnd {}
```
waynehoover, line 154: removed this fx from the base contract because it's empty and doesn't need to be here.

GarrettJMU (approved):

PR #98, opened 2023-02-13
Change Start fx to Queue

jonathandiep:
Will be helpful to have these events as the indexer can utilize them

### DIFF
--- a/contracts/Quest.sol
+++ b/contracts/Quest.sol
@@ -20,7 +20,7 @@ contract Quest is PausableUpgradeable, OwnableUpgradeable, IQuest {
     uint256 public startTime;
     uint256 public totalParticipants;
     uint256 public rewardAmountInWeiOrTokenId;
-    bool public hasQueued;
+    bool public queued;
     string public questId;
     uint256 public redeemedTokens;
 
@@ -52,7 +52,7 @@ contract Quest is PausableUpgradeable, OwnableUpgradeable, IQuest {
     /// @notice Queues the Quest
     /// @dev Only the owner of the Quest can call this function
     function queue() public virtual onlyOwner {
-        hasQueued = true;
+        queued = true;
         emit Queued(block.timestamp);
     }
 
@@ -87,13 +87,13 @@ contract Quest is PausableUpgradeable, OwnableUpgradeable, IQuest {
 
     /// @notice Checks if the Quest has started at the function level
     modifier onlyStarted() {
-        if (!hasQueued) revert NotStarted();
+        if (!queued) revert NotStarted();
         _;
     }
 
     /// @notice Checks if quest has started both at the function level and at the start time
     modifier onlyQuestActive() {
-        if (!hasQueued) revert NotStarted();
+        if (!queued) revert NotStarted();
         if (block.timestamp < startTime) revert ClaimWindowNotStarted();
         _;
     }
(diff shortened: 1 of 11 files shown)


contracts/Quest.sol
```diff
@@ -53,26 +53,31 @@ contract Quest is OwnableUpgradeable, IQuest {
...
     function unPause() external onlyOwner onlyStarted {
         isPaused = false;
+        emit Paused(false);
     }
 
     /// @notice Marks token ids as claimed
     /// @param tokenIds_ The token ids to mark as claimed
     function _setClaimed(uint256[] memory tokenIds_) private {
-        for (uint i = 0; i < tokenIds_.length;) {
+        for (uint i = 0; i < tokenIds_.length; ) {
```
JamieLottering, line 76: Unnecessary trailing whitespace here after the semicolon?

contracts/Quest.sol
```diff
@@ -105,11 +110,15 @@ contract Quest is OwnableUpgradeable, IQuest {
         if (tokens.length == 0) revert NoTokensToClaim();
 
         uint256 redeemableTokenCount = 0;
-        for (uint i = 0; i < tokens.length;) {
+        for (uint i = 0; i < tokens.length; ) {
```
JamieLottering, line 113: Same here, intentional?

contracts/interfaces/IQuest.sol
```diff
@@ -6,6 +6,8 @@ pragma solidity =0.8.16;
 interface IQuest {
     // This event is triggered whenever a call to #claim succeeds.
     event Claimed(address account, address rewardAddress, uint256 amount);
+    event Started(uint timestamp);
```
GarrettJMU, line 9: ugh this is still confusing. Wondering if we rename this to queue

JamieLottering (approved):

waynehoover:
I wonder if we should use the OZ Pausable contract, it includes events and it is a standard more or less: https://docs.openzeppelin.com/contracts/2.x/api/lifecycle#pausable

contracts/interfaces/IQuest.sol
jonathandiep: I'm cool with it, but should we rename just the event or the function as well?

contracts/Quest.sol
jonathandiep: It's intentional. I have this plugin formatting the solidity files: https://marketplace.visualstudio.com/items?itemName=NomicFoundation.hardhat-solidity

GarrettJMU (approved):

JamieLottering (approved):

contracts/Quest.sol
```diff
@@ -20,7 +20,7 @@ contract Quest is PausableUpgradeable, OwnableUpgradeable, IQuest {
     uint256 public startTime;
     uint256 public totalParticipants;
     uint256 public rewardAmountInWeiOrTokenId;
-    bool public hasStarted;
+    bool public hasQueued;
```
waynehoover, line 23: instead of `hasQueued` what about just `Queued`?

contracts/interfaces/IQuest.sol
```diff
@@ -6,6 +6,7 @@ pragma solidity =0.8.16;
 interface IQuest {
     // This event is triggered whenever a call to #claim succeeds.
     event Claimed(address account, address rewardAddress, uint256 amount);
+    event Queued(uint timestamp);
```
waynehoover, line 9: Whops just catching this now, but we need to add `indexed` to these params.

waynehoover (changes requested):

waynehoover (approved):

PR #100, opened 2023-02-13
Use OZ pausable

waynehoover:


test/Quest.spec.ts
```diff
@@ -244,7 +244,7 @@
 //
 //             await expect(deployedQuestContract.claim(firstAddress.address, objectOfAddressesAndRewards[firstAddress.address], claim.proof)).to.be.revertedWithCustomError(
 //                 questContract,
-//                 'QuestPaused'
+//                 'Pausable: paused'
 //             )
 //         })
 //
```
waynehoover, line 250: We should probably just remove this file. When we convert to an abstract contract we can add it back and test it directly then.

GarrettJMU (approved):

jonathandiep (approved):

PR #102, opened 2023-02-14
[DEV-1660] user can still claim 1155 rewards after withdrawRemainingTokens

waynehoover:


test/Erc1155Quest.spec.ts
```diff
@@ -319,6 +319,18 @@ describe('Erc1155Quest', () => {
...
+    it('should allow a user to claim after withdrawRemainingTokens is called', async () => {
+      await deployedRabbitholeReceiptContract.setMinterAddress(deployedFactoryContract.address)
+      await deployedQuestContract.start()
+      await time.setNextBlockTimestamp(startDate + 1)
+      await deployedFactoryContract.connect(firstAddress).mintReceipt(questId, messageHash, signature)
+      await time.setNextBlockTimestamp(expiryDate + 1)
+
+      await deployedQuestContract.connect(owner).withdrawRemainingTokens()
+
+      await deployedQuestContract.connect(firstAddress).claim()
```
jonathandiep, line 331: Gonna assume this is fine, but can we also have a check here for `firstAddress` token balance?

waynehoover: good idea, added.

PR #112, opened 2023-04-11
Add mintFee

waynehoover:
This adds a global mintFee to the factory contract.

### DIFF
--- a/contracts/QuestFactory.sol
+++ b/contracts/QuestFactory.sol
@@ -299,6 +299,7 @@ contract QuestFactory is Initializable, OwnableUpgradeable, AccessControlUpgrade
             if (change > 0) {
                 (bool success, ) = msg.sender.call{value: change}("");
                 require(success, "Failed to return change");
+                emit ExtraMintFeeReturned(msg.sender, change);
             }
 
             // Send the mint fee to the mint fee recipient
--- a/test/QuestFactory.spec.ts
+++ b/test/QuestFactory.spec.ts
@@ -35,6 +35,7 @@ describe('QuestFactory', () => {
   let owner: SignerWithAddress
   let royaltyRecipient: SignerWithAddress
   let protocolRecipient: SignerWithAddress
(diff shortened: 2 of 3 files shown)


contracts/QuestFactory.sol
```diff
@@ -209,13 +228,21 @@ contract QuestFactory is Initializable, OwnableUpgradeable, AccessControlUpgrade
...
     /// @param questFee_ The quest fee value
     function setQuestFee(uint16 questFee_) public onlyOwner {
         if (questFee_ > 10_000) revert QuestFeeTooHigh();
         questFee = questFee_;
     }
 
+    /// @dev set the mint fee
+    /// @notice the mint fee in ether
+    /// @param mintFee_ The mint fee value
+    function setMintFee(uint mintFee_) public onlyOwner {
```
jonathandiep, line 241: Is the mint fee priced in wei?

test/QuestFactory.spec.ts
```diff
@@ -416,5 +416,35 @@ describe('QuestFactory', () => {
...
+    it('should succeed if the mint fee is equal or greator than required amount, and only recieve the required amount', async function () {
+      const requiredFee = 1000
+      await erc20Quest.queue()
+      await deployedFactoryContract.setMintFee(requiredFee)
+      await time.setNextBlockTimestamp(startDate)
+      const balanceBefore = await ethers.provider.getBalance(deployedFactoryContract.getMintFeeRecipient())
+
+      await expect(
+        deployedFactoryContract.mintReceipt(erc20QuestId, messageHash, signature, {
+          value: requiredFee + 100,
```
jonathandiep, line 442: Since we're sending more than `requiredFee` can we also have a test to check that the refund excess works?

jonathandiep, line 442: Since we're sending more than `requiredFee` can we also have a test to check that the refund excess works?

jonathandiep (approved):

PR #113, opened 2023-04-12
Remove 1155 Quests

waynehoover:
- Combines Quest and Erc20Quest into one file
- removes Erc1155Quest
- also keeps the 1155 address storage variable in the factory mostly to keep backwards compatability

### DIFF
--- a/contracts/QuestFactory.sol
+++ b/contracts/QuestFactory.sol
@@ -31,7 +31,7 @@ contract QuestFactory is Initializable, OwnableUpgradeable, AccessControlUpgrade
     address public claimSignerAddress;
     address public protocolFeeRecipient;
     address public erc20QuestAddress;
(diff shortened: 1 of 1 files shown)


contracts/QuestFactory.sol
```diff
@@ -33,7 +31,7 @@ contract QuestFactory is Initializable, OwnableUpgradeable, AccessControlUpgrade
     address public claimSignerAddress;
     address public protocolFeeRecipient;
     address public erc20QuestAddress;
-    address public erc1155QuestAddress;
+    address public erc1155QuestAddress; // todo remove this
```
jonathandiep, line 34: Curious, how will it be possible to remove this? Do you just assign a new address variable that could overwrite this slot?

jonathandiep (approved):

PR #116, opened 2023-04-14
use cloneDeterministic when creating quests

waynehoover:
this uses the create2 opcode with a salt of msg.sender and rewardTokenAddress. This will stop re-org attacks.


contracts/QuestFactory.sol
```diff
@@ -86,7 +86,7 @@ contract QuestFactory is Initializable, OwnableUpgradeable, AccessControlUpgrade
         if (keccak256(abi.encodePacked(contractType_)) == ERC20) {
             if (rewardAllowlist[rewardTokenAddress_] == false) revert RewardNotAllowed();
 
-            address newQuest = Clones.clone(erc20QuestAddress);
+            address newQuest = Clones.cloneDeterministic(erc20QuestAddress, keccak256(abi.encodePacked(msg.sender, rewardTokenAddress_)));
```
jonathandiep, line 89: Read the docs on `cloneDeterministic` and it states "Using the same implementation and salt multiple time will revert, since the clones cannot be deployed twice at the same address."

Wouldn't it be better to use `questId_` rather than `rewardTokenAddress_`? Esp considering that the same `rewardTokenAddress_` can exist for multiple quests

PR #121, opened 2023-04-19
add createQuestAndStart fx

waynehoover:
This backwards compatible change allows for a new flow to create quests in which the tokens are moved from the quest creator to the quest and the quest is started in 1 tx.

### DIFF
--- a/contracts/QuestFactory.sol
+++ b/contracts/QuestFactory.sol
@@ -304,7 +304,7 @@ contract QuestFactory is Initializable, OwnableUpgradeable, AccessControlUpgrade
 
         Quest storage currentQuest = quests[questId_];
         if (currentQuest.numberMinted + 1 > currentQuest.totalParticipants) revert OverMaxAllowedToMint();
-        if (currentQuest.addressMinted[msg.sender] == true) revert AddressAlreadyMinted();
+        if (currentQuest.addressMinted[msg.sender]) revert AddressAlreadyMinted();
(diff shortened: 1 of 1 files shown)


contracts/QuestFactory.sol
```diff
@@ -115,7 +113,85 @@ contract QuestFactory is Initializable, OwnableUpgradeable, AccessControlUpgrade
...
+            rewardTokenAddress_,
+            endTime_,
+            startTime_,
+            totalParticipants_,
+            rewardAmount_,
+            contractType_,
+            questId_
+        );
+
+        transferTokensAndQueueQuest(newQuest, rewardTokenAddress_);
```
jonathandiep, line 192: Should we have a modifier/check to see if the QuestFactory is approved to transfer the amount of reward tokens?

contracts/Quest.sol
```diff
@@ -66,10 +66,15 @@ contract Quest is ReentrancyGuardUpgradeable, PausableUpgradeable, OwnableUpgrad
         __ReentrancyGuard_init();
     }
 
+    /// @dev The amount of tokens the quest needs to pay all redeemers plus the protocol fee
+    function totalTransferAmount() external view returns (uint256) {
+        return this.maxTotalRewards() + this.maxProtocolReward();
+    }
```
jonathandiep, line 72: I love that you exported this out into its own fx. Easier to just call this than having to calculate all the time

PR #123, opened 2023-05-02
add QuestTerminalKey NFT

waynehoover:


test/Quest.spec.ts
```diff
@@ -95,6 +95,7 @@ describe('Erc20Quest', async () => {
       protocolFeeRecipient.address,
       deployedErc20Quest.address,
       owner.address,
+      owner.address, // as a placeholder, would the QuestTerminalKey contract
```
jonathandiep, line 98: Should we also have `deployQuestTerminalKey` in this spec file as well in order to update `questTerminalKeyContract` variable?

[nit] Maybe zero address would be a better placeholder

contracts/QuestTerminalKey.sol
```diff
@@ -0,0 +1,192 @@
...
+        emit RoyaltyFeeSet(royaltyFee_);
+    }
+
+    /// @dev mint a QuestTerminalKey NFT
+    /// @param to_ the address to mint to
+    /// @param discountPercentage_ the discount percentage
+    /// @param maxDiscountUses_ the max discount uses
+    function mint(address to_, uint16 discountPercentage_, uint16 maxDiscountUses_) external onlyMinter {
+        require(discountPercentage_ <= 10000, 'Invalid discount percentage');
+        require(maxDiscountUses_ <= 1000, 'Invalid max discount uses');
```
jonathandiep, line 113: So the minimum amount of times the nft can use the discount is 1001? Is there a reason for that?

jonathandiep: Oops sorry read this backwards LOL. It requires 1000 or less

I guess we want to set an upper limit for the amount of times the discount can be used

contracts/QuestTerminalKey.sol
```diff
@@ -0,0 +1,192 @@
...
+        _;
+    }
+
+    function _baseURI() internal pure override returns (string memory) {
+        return "https://api.rabbithole.gg/nft/qtk/";
+    }
+
+    /// @dev modifier to check for zero address
+    /// @param _address the address to check
+    modifier nonZeroAddress(address _address) {
```
jonathandiep, line 75: Is there any benefit for this modifier? Does/can going without it save gas? (imo the check is more important than saving marginal amount of gas)

contracts/QuestFactory.sol
```diff
@@ -111,13 +123,25 @@ contract QuestFactory is Initializable, OwnableUpgradeable, AccessControlUpgrade
...
         return newQuest;
     }
 
+    // do we require or return questfee when maxUses are up? -> currently return qusetfee
+    // will another person be creating quests on behalf of the owner of the discount token? if so the require below will need to change -> currently not possible
+    function doDiscountedFee(uint tokenId_) internal returns (uint16) {
+        require(questTerminalKeyContract.ownerOf(tokenId_) == msg.sender, "QuestFactory: caller is not owner of discount token");
+
+        (uint16 discountPercentage, uint16 maxDiscountUses, uint16 usedCount) = questTerminalKeyContract.discounts(tokenId_);
+        if(usedCount > maxDiscountUses) return questFee;
```
jonathandiep, line 139: I think this might need to be `>=` (please verify!)

Lets say we set `maxUses` to 1 when we mint the QuestTerminalKey NFT
- We deploy a quest -- now `usedCount` is set to 1
- We can still deploy another discounted quest because `usedCount` is 1 and `maxDiscountUses` is also 1

test/QuestFactory.spec.ts
```diff
@@ -211,7 +231,44 @@ describe('QuestFactory', () => {
         totalRewards,
         rewardAmount,
         'erc20',
-        erc20QuestId
+        erc20QuestId,
+        0 //TODO test with a created discountTokenId
```
jonathandiep, line 235: Can remove this TODO

test/QuestFactory.spec.ts
jonathandiep, line 1: We should include a test where we exceed `maxUses` and expect the quest fee without discounts

test/QuestFactory.spec.ts
```diff
@@ -211,7 +231,44 @@ describe('QuestFactory', () => {
...
+
+      const tx = await deployedFactoryContract.createQuestAndQueue(
+        deployedSampleErc20Contract.address,
+        expiryDate,
+        startDate,
+        totalRewards,
+        rewardAmount,
+        'erc20',
+        erc20QuestId,
+        discountTokenId
```
jonathandiep, line 271: We should add a check here to make sure quest fee is what we expect it to be

test/QuestTerminalKey.spec.ts
```diff
@@ -0,0 +1,81 @@
...
+      expect(await questTerminalKey.royaltyRecipient()).to.equal(royaltyRecipient.address)
+    })
+  })
+
+  describe('mint', () => {
+    it('mints a token with correct questId', async () => {
+      await questTerminalKey.connect(minterAddress).mint(firstAddress.address, 100, 3)
+      const discount = await questTerminalKey.discounts(1)
+
+      expect(await questTerminalKey.balanceOf(firstAddress.address)).to.eq(1)
```
jonathandiep, line 55: Maybe another simple check to add here is to verify that `ownerOf` `tokenId` 1 is the address of `firstAddress`

contracts/QuestTerminalKey.sol
```diff
@@ -0,0 +1,243 @@
...
+            ']'
+        );
+        bytes memory dataURI = abi.encodePacked(
+            '{',
+            '"name": "RabbitHole.gg QuestTerminalKey #',
+            tokenIdString,
+            '",',
+            '"description": "The RabbitHole.gg QuestTerminalKey is used as key to access the Terminal.",',
+            '"image": "',
+            "https://rabbithole.gg/_next/image?url=%2FQTKNFT.png",
```
jonathandiep, line 197: Is this how we will host the image? Wondering if we should use IPFS here

contracts/QuestTerminalKey.sol
```diff
@@ -159,14 +160,64 @@ contract QuestTerminalKey is
...
+    /// @dev returns the data uri in json format
+    /// @param tokenId_ the token id
+    function generateDataURI(
+        uint tokenId_
+    ) internal view virtual returns (bytes memory) {
+        string memory tokenIdString = tokenId_.toString();
+
+        bytes memory attributes = abi.encodePacked(
+            '[',
+            generateAttribute('Discount Percentage', discounts[tokenId_].percentage.toString()),
```
jonathandiep: Will this display the bips value?

jonathandiep: Nvm I see you updated the attribute key (I was reviewing old commit here)

jonathandiep (approved):

PR #124, opened 2023-05-04
add jsonSpecCID to createQuestAndQueue

waynehoover:


jimobrien (approved):

contracts/Quest.sol
```diff
@@ -66,6 +67,15 @@ contract Quest is ReentrancyGuardUpgradeable, PausableUpgradeable, OwnableUpgrad
         __ReentrancyGuard_init();
     }
 
+    /// @dev set jsonSpecCID only if its empty
+    /// @param jsonSpecCID_ The jsonSpecCID to set
+    function setJsonSpecCID(string memory jsonSpecCID_) external onlyOwner {
+        require(bytes(jsonSpecCID_).length > 0, 'jsonSpecCID cannot be empty');
+        require(bytes(jsonSpecCID).length == 0, 'jsonSpecCID already set');
```
jonathandiep, line 74: So once set, we cannot edit?

jonathandiep: I guess I kinda agree with this. So owner can't change requirements in middle of quest

Wondering if we can allow edits until the quest starts?

contracts/Quest.sol
```diff
@@ -66,6 +67,15 @@ contract Quest is ReentrancyGuardUpgradeable, PausableUpgradeable, OwnableUpgrad
...
+
+        jsonSpecCID = jsonSpecCID_;
```
jonathandiep, line 76: Should we emit event?

jonathandiep (approved):
Left a couple comments, but lgtm!

PR #126, opened 2023-05-11
add QuestNFT contract and createQuestNFT and mintQuestNFT fxs on QuestFactory

waynehoover:


contracts/QuestNFT.sol
```diff
@@ -0,0 +1,219 @@
...
+        internal
+        whenNotPaused
+        override
+    {
+        super._beforeTokenTransfer(from, to, tokenId, batchSize);
+    }
+
+    /// @dev Function that to withdraw the remaining coins in the contract to the owner
+    /// @notice This function can only be called after the quest end time
+    function withdrawRemainingTokens() external onlyAfterQuestEnd {
```
jonathandiep, line 138: Should we change function name since it doesn't transfer tokens?

contracts/QuestNFT.sol
```diff
@@ -0,0 +1,219 @@
...
+        _pause();
+    }
+
+    /// @notice Unpauses the Quest
+    /// @dev Only the owner of the Quest can call this function.
+    function unPause() external onlyOwner {
+        _unpause();
+    }
+
+    function safeMint(address to_) public onlyMinter onlyQuestBetweenStartEnd whenNotPaused nonReentrant {
```
jonathandiep, line 120: Do we want to cap the amount of mints? Maybe to the value of `totalParticipants`?

contracts/QuestNFT.sol
```diff
@@ -0,0 +1,219 @@
...
+        uint256 questFee_,
+        address protocolFeeRecipient_,
+        address minterAddress_, // should always be the QuestFactory contract
+        string memory jsonSpecCID_,
+        string memory name_,
+        string memory symbol_,
+        string memory description_,
+        string memory imageIPFSHash_
+    ) external initializer {
+        require (endTime_ > block.timestamp, 'endTime_ in past');
```
jonathandiep, line 56: Extremely extremely rare edge case, but if `block.timestamp` == `endTime_` the error message is a little inaccurate

contracts/QuestNFT.sol
```diff
@@ -0,0 +1,219 @@
...
+    /// @dev Only the owner of the Quest can call this function.
+    function unPause() external onlyOwner {
+        _unpause();
+    }
+
+    function safeMint(address to_) public onlyMinter onlyQuestBetweenStartEnd whenNotPaused nonReentrant {
+        _tokenIdCounter.increment();
+        uint tokenId = _tokenIdCounter.current();
+        _safeMint(to_, tokenId);
+        (bool success, ) = protocolFeeRecipient.call{value: questFee}("");
```
jonathandiep, line 124: Noob question, what does this line do? Does it transfer `questFee` ETH from `msg.sender` to `protocolFeeRecipient`?

contracts/QuestNFT.sol
```diff
@@ -0,0 +1,219 @@
...
+            '"image": "',
+            tokenImage(),
+            '"',
+            '}'
+        );
+        return dataURI;
+    }
+
+    function tokenImage() internal view virtual returns (string memory) {
+        return string(abi.encodePacked('ipfs://', imageIPFSHash));
```
jonathandiep, line 185: Should we have a `setTokenImage` function? Or is intention for it to be permanent?

contracts/QuestFactory.sol
```diff
@@ -215,6 +237,93 @@ contract QuestFactory is Initializable, OwnableUpgradeable, AccessControlUpgrade
...
+            protocolFeeRecipient,
+            address(this),
+            data.jsonSpecCID,
+            data.name,
+            data.symbol,
+            data.description,
+            data.imageIPFSHash
+        );
+
+        (bool success, ) = payable(newQuest).call{value: msg.value}("");
```
jonathandiep, line 291: What is the purpose of having this here? Could it also be called if `msg.value` is set to 0?

jonathandiep: Ahh requires `msg.value` >= `totalQuestNFTFee` when calling `createQuestNFTInternal()`

jonathandiep (approved):

PR #128, opened 2023-05-18
remove start time check from Quest and add animation_url metadata to QuestTerminalKey

waynehoover:


contracts/Quest.sol
```diff
@@ -49,7 +49,6 @@ contract Quest is ReentrancyGuardUpgradeable, PausableUpgradeable, OwnableUpgrad
         address protocolFeeRecipient_
     ) external initializer {
         if (endTime_ <= block.timestamp) revert EndTimeInPast();
-        if (startTime_ <= block.timestamp) revert StartTimeInPast();
```
jonathandiep, line 52: I think we're gonna need to remove this test?

Can we also remove the `error StartTimeInPast()` from the interface? Or are we going to need to keep for upgradability?

jimobrien, line 52: 🙇

jimobrien (approved):

test/QuestFactory.spec.ts
```diff
@@ -67,6 +67,7 @@ describe('QuestFactory', () => {
       10,
       owner.address,
       'QmTy8w65yBXgyfG2ZBg5TrfB2hPjrDQH3RCQFJGkARStJb',
+      'QmTy8w65yBXgyfG2ZBg5TrfB2hPjrDQH3RCQFJGkARzzzz',
```
jonathandiep, line 70: [nit] can we use [QmcniBv7UQ4gGPQQW2BwbD4ZZHzN3o3tPuNLZCbBchd1zh](https://ipfs.io/ipfs/QmcniBv7UQ4gGPQQW2BwbD4ZZHzN3o3tPuNLZCbBchd1zh) for the `animation_url`

jonathandiep (approved):

PR #141, opened 2023-06-16
Move questNFT to a 1155

waynehoover:
This changes the NFTQuest type from an ERC721 to an ERC1155 type. When an erc1155 is created we call this a collection. A collection can have many quests which map to token ids in the 1155. The flow is this:

Quest creator:
- Deploy a new erc1155 "collection" `QuestFactory.createCollection`
- Add an NFT to the collection `QuestFactory.addQuestToCollection`

Quest completer:
- mint an NFT from the collection `QuestFactory.mintQuestNFT`

And we have some handy functions for the frontend to use:
`QuestFactory.ownerCollectionsByOwner` returns all the collections given an user address. This will be useful because a user can choose to either use an existing collection or create a new one when creating a NFT type quest.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ For more information on all docs related to the Quest Protocol, see the document
 
 ---
 
-## Addressses
+## Addresses
 
 Mainnet, Optimism, Polygon, Arbitrum, and Sepolia:
 
--- a/docs/overview.md
+++ b/docs/overview.md
@@ -11,14 +11,14 @@ can use the `Receipt` to claim the `Reward` in the future and any other potentia
 ## Key Concepts
 
 - **Receipts**: An NFT (ERC-721) representing a receipt from completing an action defined in the `Quest`. These are
-  originally minted to a participants EOA address and the amount of available Receipts are defined by the `Quest`
+  originally minted to a participants EOA address and the amount of available Receipts is defined by the `Quest`
   contract. Receipts track ability to see if there is a reward to claim.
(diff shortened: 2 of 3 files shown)


contracts/QuestNFT.sol
```diff
@@ -15,73 +16,53 @@ import {ReentrancyGuardUpgradeable} from "@openzeppelin/contracts-upgradeable/se
 /// @title QuestNFT
 /// @author RabbitHole.gg
 /// @notice This contract is the Erc721 Quest Completion contract. It is the NFT that can be minted after a quest is completed.
```
jonathandiep, line 18: Should update this comment to use 1155

contracts/QuestNFT.sol
```diff
@@ -90,19 +71,54 @@ contract QuestNFT is Initializable, ERC721Upgradeable, PausableUpgradeable, Owna
...
+        uint256 questFee_,
+        uint256 startTime_,
+        uint256 endTime_,
+        uint256 totalParticipants_,
+        string memory questId_,
+        string memory description_,
+        string memory imageIPFSHash_
+    ) public onlyMinter returns (uint256) {
+        require (endTime_ > block.timestamp, 'endTime_ in the past');
+        require (startTime_ > block.timestamp, 'startTime_ in the past');
```
jonathandiep, line 84: Should we remove this check? I believe we removed it in `Quest.sol` so that we can start quests immediately

jonathandiep (approved):
Left a couple of comments, but lgtm!

PR #149, opened 2023-07-20
Share To Earn

waynehoover:
- new function `claim` 
  - this is a universal function that calls the correct erc1155 or erc20 claim function, it also takes a referral address which should be set to `0` when no referral is present.
- new public storage var `referralFee` which is the percent (in bips) of the mintFee given to affiliates
  - has onlyOwner set function


contracts/QuestFactory.sol
```diff
@@ -549,20 +602,33 @@ contract QuestFactory is Initializable, OwnableUpgradeable, AccessControlUpgrade
         ++currentQuest.numberMinted;
         questContract_.singleClaim(msg.sender);
 
-        if(mintFee > 0) processMintFee();
+        if(mintFee > 0) processMintFee(ref_);
 
         emit Quest1155Claimed(msg.sender, currentQuest.questAddress, questId_, questContract_.rewardToken(), questContract_.tokenId());
+
+        if(ref_ != address(0)) {
+            emit QuestClaimedReferred(msg.sender, currentQuest.questAddress, questId_, questContract_.rewardToken(), questContract_.tokenId(), ref_, referralFee);
```
jonathandiep, line 610: In the case that this does get emitted, do we want to emit both `Quest1155Claimed` and `QuestClaimedReferred`?

contracts/QuestFactory.sol
```diff
@@ -549,20 +602,33 @@ contract QuestFactory is Initializable, OwnableUpgradeable, AccessControlUpgrade
...
+            getMintFeeRecipient().safeTransferETH(mintFee);
+            return;
+        }
+        uint referralAmount = (mintFee * referralFee) / 10_000;
+        ref_.safeTransferETH(referralAmount);
+        getMintFeeRecipient().safeTransferETH(mintFee - referralAmount);
     }
 
-    function processMintFee() private {
+    function processExtraFee() private {
```
jonathandiep, line 625: [nit] extra fees sound like it's going to be actual extra fees lol. Maybe wondering if we can change to `processChange` or `processRefund` or something

contracts/QuestFactory.sol
```diff
@@ -529,16 +570,28 @@ contract QuestFactory is Initializable, OwnableUpgradeable, AccessControlUpgrade
         ++currentQuest.numberMinted;
         questContract_.singleClaim(msg.sender);
 
-        if(mintFee > 0) processMintFee();
+        if(mintFee > 0) processMintFee(ref_);
 
         emit QuestClaimed(msg.sender, currentQuest.questAddress, questId_, questContract_.rewardToken(), questContract_.rewardAmountInWei());
+
+        if(ref_ != address(0)) {
+            emit QuestClaimedReferred(msg.sender, currentQuest.questAddress, questId_, questContract_.rewardToken(), questContract_.rewardAmountInWei(), ref_, referralFee);
```
jonathandiep, line 578: Same question here that I asked earlier: In the case that this does get emitted, do we want to emit both `QuestClaimed` and `QuestClaimedReferred`?

contracts/QuestFactory.sol
jonathandiep: Would we also want to use the same event type for both erc20 and erc1155 quests? Since both get emitted from QuestFactory, might be difficult telling which type of quest this event refers to (though it does have `questId`, but just thinking here)

jonathandiep (approved):

PR #150, opened 2023-07-20
add allowlist for nftQuestFee

waynehoover:
- Adds `NFTQuestFeeList` mapping to map nft quest fees to addresses
- uses mapping when creating an NFTQuest for the nftQuestFee

### DIFF
--- a/contracts/QuestFactory.sol
+++ b/contracts/QuestFactory.sol
@@ -129,8 +129,8 @@ contract QuestFactory is Initializable, OwnableUpgradeable, AccessControlUpgrade
         _;
     }
 
-    modifier nonZeroAddress(address _address) {
-        if (_address == address(0)) revert ZeroAddressNotAllowed();
+    modifier nonZeroAddress(address address_) {
+        if (address_ == address(0)) revert ZeroAddressNotAllowed();
         _;
     }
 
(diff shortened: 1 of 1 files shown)


contracts/QuestFactory.sol
```diff
@@ -556,6 +565,13 @@ contract QuestFactory is Initializable, OwnableUpgradeable, AccessControlUpgrade
         getMintFeeRecipient().safeTransferETH(mintFee);
     }
 
+    function setNFTQuestFeeList(address[] calldata toAddAddresses, uint[] calldata fees) external onlyOwner
+    {
+        for (uint i = 0; i < toAddAddresses.length; i++) {
+            NFTQuestFeeList[toAddAddresses[i]] = NFTQuestFees(fees[i], true);
```
jonathandiep, line 571: Should we emit event for this? My reasoning is that it would be easier to track onchain for fee set activity

test/QuestFactory.spec.ts
```diff
@@ -600,5 +600,35 @@ describe('QuestFactory', () => {
 
       expect(await deployedSampleErc1155Contract.balanceOf(questUser.address, NFTTokenId)).to.equal(1)
     })
+
+    it('Should claim rewards from a 1155 Quest with zero questFee', async () => {
```
jonathandiep, line 604: [nit] can we change title of "zero questFee" to "zero nftQuestFee"?

jonathandiep (approved):

PR #176, opened 2023-08-17
remove sablier stream cancellation

waynehoover:
Since we aren't planning to cancel streams we can leave this as default false.


Quazia (approved):
LGTM

Issue #175, opened 2023-08-17
Push comments/docs into interfaces


Quazia:
The code will look cleaner if we push all our comments and doc strings into the contract interfaces and just leverage the `@inheritdoc` tag. I'd do this after we sort them, we could also sort the interface functions alphabetically, but I'm not strongly opinionated on that. I think after visibility sorting either functionality sorting, or alphabetical sorting make equal sense.

Should be handled after #174

Issue #172, opened 2023-08-16
Standardize to Custom Errors


Quazia:
Right now we're using mixed errors, some `string` based `require` guards, and other `custom error` based assert guards. I think it's generally more efficient to use custom errors, and I love the ability to drop params into them easily during debugging. But I'm open to standardizing to either format.

PR #177, opened 2023-08-17
🎨QuestFactory&Quest: use interfaces vs instances

### LINKED ISSUE #168: Only use interfaces in QuestFactory
Instead of importing Quest.sol and Quest1155.sol in QuestFactory let's just use the interfaces instead. Look at https://github.com/rabbitholegg/quest-protocol/pull/152/commits/52b682ee23c438be158b1dcc74b284c7b10c60c2#diff-46d8d9ffae356f90bd43e36eef9e04ca41262e025e815f70e7434b80412bdf4e and the commits after for inspiration.

Quazia:
This PR sorts out the import statements in two of our highest traffic contracts, `QuestFactory` and `Quest` then incrementally moves each of them away from using direct contract instances, towards using contract interfaces. I also added some interfaces that were missing and cleanup some of the interfaces that I was touching. Not all interfaces have gotten this update yet, and I'll handle the rest of this when I handle #175.

I made a few aggregate interfaces. Alternatively, I could have had the interface inherit from the secondary interface and handle the overrides in the implementing interface (`IQuest` inherits from `IOwnable`) but I like this aggregate approach since it leaves a little more flexibility. If we prefer to collapse all of the functionality into the one main interface over using interface aggregates I'd be happy to make that change.

I also started the process of pulling events, structs, and errors into the interfaces but that work will continue in #175 and #172. For now I only did that in interfaces I was already touching for this PR.

Closes #168

### DIFF
--- a/contracts/Quest.sol
+++ b/contracts/Quest.sol
@@ -1,15 +1,18 @@
 // SPDX-License-Identifier: UNLICENSED
 pragma solidity ^0.8.18;
 
+// Inherits
+import {Ownable} from 'solady/src/auth/Ownable.sol';
 import {PausableUpgradeable} from '@openzeppelin/contracts-upgradeable/security/PausableUpgradeable.sol';
 import {ReentrancyGuardUpgradeable} from '@openzeppelin/contracts-upgradeable/security/ReentrancyGuardUpgradeable.sol';
-import {Ownable} from 'solady/src/auth/Ownable.sol';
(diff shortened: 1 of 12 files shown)


Quazia:
I have a commit ready downstream to fix the `solhint` issues (empty blocks) but I'm keeping it downstream just so you can see what it's doing with the in-line warnings. I think I could get it to actually block merging on warnings if that's the direction we wanna go.

PR #184, opened 2023-08-18
🔨forge: coverage and gas reporting

### LINKED ISSUE #178: Setup coverage testing
We can use `Forge coverage` to generate coverage reports and `Coverage gutters` in VSCode to highlight uncovered lines in IDE.

Quazia:
Not sure when we'll want to bring this into main, but I doubt this PR will go stale. I left the `hardhat` coverage command and gas reporting here for the time being - although it doesn't look like the `hardhat coverage` is working right now. Is it worth trying to get that up and running for the migration towards `foundry`? Also want to automate the coverage badge still.

closes #178


Quazia:
Is this okay to bring in still? I don't see any reason not to now that there's no HH testing

Quazia:
Oh we're not using any package scripts anymore so I should probably remove that part...

PR #195, opened 2023-08-31
use defaultMintFeeRecipient when passing zero address

waynehoover:


script/ProtocolRewards.s.sol
```diff
@@ -39,6 +39,21 @@ contract ProtocolRewardsUpgrade is Script {
...
 
+        vm.stopBroadcast();
+    }
+}
+
+
+contract ProtocolRewardsUpdateRole is Script {
+    function run() external {
+        uint256 deployerPrivateKey = vm.envUint("MAINNET_PRIVATE_KEY");
+        address payable protocolRewardsProxy = payable(0x168437d131f8deF2d94B555FF34f4539458DD6F9);
```
Quazia, line 50: Preference for using named constants vs inline addresses for readability

Quazia (approved):
That re-use of the deprecated slot is a neat trick.

I'd make a single file with the constants and then just import/consume that for readability but otherwise looks clean

Issue #196, opened 2023-09-01
Remove all hardhat scripts


waynehoover:
Need to double-check, but all of our hardhat scripts aren't used anymore and have been replaced by forge. We can remove them.

PR #198, opened 2023-09-05
remove hardhat scripts and add comments

waynehoover:
- remove hardhat scripts closes https://github.com/rabbitholegg/quest-protocol/issues/196
- remove hardhat deploy, and associated deploy scripts
- add comments to various places they were missing

### DIFF
--- a/test/QuestFactory.spec.ts
+++ b/test/QuestFactory.spec.ts
@@ -499,19 +499,13 @@ describe('QuestFactory', () => {
         deployedFactoryContract.connect(questUser).claimRewards(erc20QuestId, messageHash, signature, {
           value: requiredFee + extraChange,
         })
-      ).to.emit(deployedFactoryContract, 'QuestClaimed')
-        .to.emit(deployedFactoryContract, 'MintFeePaid')
+      )
+        .to.emit(deployedFactoryContract, 'QuestClaimed')
         .to.emit(deployedFactoryContract, 'ExtraMintFeeReturned')
         .withArgs(questUser.address, extraChange)
-
       expect(await ethers.provider.getBalance(deployedFactoryContract.getMintFeeRecipient(owner.address))).to.equal(
-        balanceBefore.add(requiredFee / 2)
+        balanceBefore.add(requiredFee)
       )
-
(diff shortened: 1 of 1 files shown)


contracts/QuestFactory.sol
```diff
@@ -541,24 +541,11 @@ contract QuestFactory is Initializable, OwnableUpgradeable, AccessControlUpgrade
         ++currentQuest.numberMinted;
         questContract_.singleClaim(msg.sender);
 
-        if (mintFee > 0) processMintFee(ref_, currentQuest.mintFeeRecipient);
+        if (mintFee > 0) processMintFee(ref_, currentQuest.questCreator);
```
Quazia, line 544: This struct param is unused now - if the recipient is always the creator we can just remove the setter for that in the quest create. I also think we can remove 
```
        if (mintFeeRecipient_ == address(0)) {
            mintFeeRecipient_ = defaultMintFeeRecipient;
        }
        ```
        Since creator is never 0 but worth double checking that...

**Review by Quazia — APPROVED** [review]
LGTM - there are some optimizations we could make as a side effect of the [this loc](https://github.com/rabbitholegg/quest-protocol/pull/198/commits/068b0dce0a5a5ed77325d7bc41ade81bbf7f27e9#diff-46d8d9ffae356f90bd43e36eef9e04ca41262e025e815f70e7434b80412bdf4eR544) but otherwise looks good. Also if we don't want to push deploy scripts maybe we git ignore that whole folder? Idk not overly opinionated. I can see having at least the recent ones be super useful.

File: contracts/interfaces/IQuestFactory.sol [hunk 2]
```diff
@@ -84,6 +84,7 @@ interface IQuestFactory {
         uint16 referralFee,
         uint256 mintFeeEthWei
     );
+    event MintFeePaid(address referrer, uint256 referrerAmount, address mintFeeRecipient, uint256 mintFeeRecipientAmount);
```
kmairin, line 87: let's do `referrerAmountWei` and `mintFeeRecipientAmountWei`

kmairin: I think questId will be needed in order for us to track earning per quest per creator

kmairin (approved):

PR #199, opened 2023-09-07
split mint fee with protocol, quest creator, and referrer

waynehoover:
- new mintFee distribution, equal parts go to the protocol quest creator and referrer. 
- add new MintFeePaid event

### DIFF
--- a/contracts/interfaces/IQuestFactory.sol
+++ b/contracts/interfaces/IQuestFactory.sol
@@ -85,7 +85,7 @@ interface IQuestFactory {
         uint256 mintFeeEthWei
     );
     event MintFeePaid(
-        string indexed questId,
+        string questId,
         address rabbitHoleAddress,
         uint256 rabbitHoleAmountWei,
         address questCreatorAddress,


contracts/QuestFactory.sol
```diff
@@ -672,17 +670,18 @@ contract QuestFactory is Initializable, OwnableUpgradeable, AccessControlUpgrade
...
-    function processMintFee(address ref_, address _mintFeeRecipient) private {
+    function processMintFee(address ref_, address mintFeeRecipient_, string memory questId_) private {
         returnChange();
-        if (_mintFeeRecipient == address(0)) {
-            _mintFeeRecipient = defaultMintFeeRecipient;
-        }
-        if (ref_ == address(0)) {
-            ref_ = defaultReferralFeeRecipient;
-        }
+        if (mintFeeRecipient_ == address(0)) mintFeeRecipient_ = defaultMintFeeRecipient;
```
Quazia, line 675: This is temporary to handle the legacy quests that don't have a creator set?

Quazia (approved):
LGTM
Once the existing quests are processed I think we can [remove this line
](https://github.com/rabbitholegg/quest-protocol/pull/199/files#diff-46d8d9ffae356f90bd43e36eef9e04ca41262e025e815f70e7434b80412bdf4eR675)

PR #200, opened 2023-09-07
split protocol fee 50% with quest creator

waynehoover:
Half of the protocol fee goes to the protocolFeeRecipient and the other half plus any remaining balance go back to the creator.


Quazia (approved):
LGTM - I like the removal of `onlyProtocolFeeRecipientOrOwner` check to allow for EOA and shave some gas.

contracts/Quest.sol
```diff
@@ -151,13 +151,19 @@ contract Quest is ReentrancyGuardUpgradeable, PausableUpgradeable, Ownable, IQue
...
-        rewardToken.safeTransfer(protocolFeeRecipient, this.protocolFee());
-        rewardToken.safeTransfer(owner(), rewardToken.balanceOf(address(this)));
+
+        uint256 protocolFeeForRecipient = this.protocolFee() / 2;
+        rewardToken.safeTransfer(protocolFeeRecipient, protocolFeeForRecipient);
+
+        uint256 remainingBalanceForOwner = rewardToken.balanceOf(address(this));
+        rewardToken.safeTransfer(owner(), remainingBalanceForOwner);
+
+        emit ProtocolFeeDistributed(questId, rewardToken, protocolFeeRecipient, protocolFeeForRecipient, owner(), remainingBalanceForOwner);
```
kmairin, line 166: [nit]Should we rename these variables to match the event signatures too?

kmairin (approved):
Lgtm with just a nit

PR #208, opened 2023-09-16
add scripts to deploy contracts

### LINKED ISSUE #160: Create foundry deploy scripts for Proxy Admin, Factory, and RabbitHoleTickets
We currently have to checkout a specific sha and use hardhat-deploy from that sha when we want to deploy our contracts to a new network. We could copy the raw bytecode for the deployments to a forge script so we don't have to do this.

We need scripts to deploy 3 contracts:
- ProxyAdmin to `0xD28fbF7569f31877922cDc31a1A5B3C504E8faa1`
- QuestFactory to `0x52629961F71C1C2564C5aa22372CB1b9fa9EBA3E`
- RabbitHoleTickets to `0x0D380362762B0cf375227037f2217f59A4eC4b9E`

It doesn't look like this can be done with forge scripts yet, but we can use forge cast:
https://book.getfoundry.sh/reference/cast/cast-send

Forge cast with the same raw data from the tx that created those contracts should work for example the ProxyAdmin on mainnet was created in this tx: https://etherscan.io/tx/0x906aeb77f2ee06382c13aa7be4c27e05e30238f3941fa0202798091875943a4c

waynehoover:
closes #160


Quazia (approved):
I like the use of `QuestContractConstants.sol` for consts - it'd be nice to save some of the initialize params somewhere so it's more clear what they are/we can re-use them in various places to make deployment changes easier in the future.
I think we could even [store some of these params in JSON](https://book.getfoundry.sh/cheatcodes/parse-json#how-to-use-stdjson) files then [import them straight into the scripts](https://book.getfoundry.sh/cheatcodes/fs?highlight=readFile#signature). The read/write might be useful to store the info into json files in addition to the broadcasts, or we could even just read the values from the broadcasts. Just some ideas. Overall looks good, just hard to read in current form.

PR #209, opened 2023-09-19
add Quest1155.t.sol

waynehoover:
- Adds Quest1155.t.sol

### DIFF
--- a/test/Quest1155.t.sol
+++ b/test/Quest1155.t.sol
@@ -22,6 +22,8 @@ contract TestQuest1155 is Test, Errors, Events {
     uint256 TOTAL_PARTICIPANTS = 300;
     uint256 TOKEN_ID = 1;
     uint16 QUEST_FEE = 2000; // 20%
+    uint256 MINT_AMOUNT = 100_000;
+    uint256 LARGE_ETH_AMOUNT = 100_000_000;
     address protocolFeeRecipient = makeAddr("protocolFeeRecipient");
     address questFactoryMock;
     address participant = makeAddr(("participant"));
@@ -142,7 +144,7 @@ contract TestQuest1155 is Test, Errors, Events {
     }
 
     function test_RevertIf_not_enough_eth() public {
-        SampleERC1155(sampleERC1155).mintSingle(address(quest), 1, 100000);
+        SampleERC1155(sampleERC1155).mintSingle(address(quest), 1, MINT_AMOUNT);
 
         vm.expectRevert(abi.encodeWithSelector(InsufficientETHBalance.selector));
         vm.prank(questFactoryMock);
@@ -150,8 +152,8 @@ contract TestQuest1155 is Test, Errors, Events {
     }
 
     function test_queue() public {
-        SampleERC1155(sampleERC1155).mintSingle(address(quest), 1, 100000);
-        vm.deal(address(quest), 100000000);
+        SampleERC1155(sampleERC1155).mintSingle(address(quest), 1, MINT_AMOUNT);
+        vm.deal(address(quest), LARGE_ETH_AMOUNT);
 
         vm.prank(questFactoryMock);
         quest.queue();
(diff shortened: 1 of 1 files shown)


test/Quest1155.t.sol
```diff
@@ -0,0 +1,306 @@
...
+            START_TIME,
+            TOTAL_PARTICIPANTS,
+            TOKEN_ID,
+            QUEST_FEE,
+            protocolFeeRecipient
+        );
+    }
+
+    function test_RevertIf_initialize_EndTimeLessThanOrEqualToStartTime() public {
+        vm.warp(START_TIME - 1); // reset time
```
Quazia, line 85: I don't think this is needed - I believe it'll reset to the startup state for each test. Not positive though.

test/Quest1155.t.sol
```diff
@@ -0,0 +1,306 @@
...
+
+        assertEq(
+            protocolFeeRecipient.balance,
+            protocolFeeRecipientOGBalance + QUEST_FEE,
+            "participant should have received the reward"
+        );
+        assertEq(
+            SampleERC1155(sampleERC1155).balanceOf(participant, TOKEN_ID),
+            1,
+            "participant should have received the reward"
```
Quazia, line 188: I'd differentiate these messages. (other being on line 182)

test/Quest1155.t.sol
```diff
@@ -0,0 +1,306 @@
...
+        );
+    }
+
+    // todo add fuzz test
```
Quazia, line 192: Probably less important with the 1155 quests but agreed. Good to have.

test/Quest1155.t.sol
```diff
@@ -0,0 +1,306 @@
...
+
+    function test_RevertIf_singleClaim_whenNotPaused() public {
+        SampleERC1155(sampleERC1155).mintSingle(address(quest), 1, 100000);
+        vm.deal(address(quest), 100000000);
+        vm.prank(questFactoryMock);
+        quest.queue();
+        vm.startPrank(questFactoryMock);
+        quest.pause();
+        vm.warp(START_TIME);
+        vm.expectRevert("Pausable: paused");
```
Quazia, line 224: Pesky OZ error messages

test/Quest1155.t.sol
```diff
@@ -0,0 +1,306 @@
...
+        vm.stopPrank();
+    }
+
+    // /*//////////////////////////////////////////////////////////////
+    //                   WITHDRAWREMAININGTOKENS
+    // //////////////////////////////////////////////////////////////*/
+
+    function test_withdrawRemainingTokens() public {
+        SampleERC1155(sampleERC1155).mintSingle(address(quest), 1, 100000);
+        vm.deal(address(quest), 100000000);
```
Quazia, line 235: If possible it'd be good to use a const vs a magic number here. Makes changing the test variable in the future easier and makes it a lot lower cognitive load to create the fuzz tests since everything is already setup deterministically.

test/Quest1155.t.sol
```diff
@@ -0,0 +1,306 @@
...
+
+    function test_RevertIf_withdrawRemainingToken_NotEnded() public {
+        SampleERC1155(sampleERC1155).mintSingle(address(quest), 1, 100000);
+        vm.deal(address(quest), 100000000);
+        vm.prank(questFactoryMock);
+        quest.transferOwnership(owner);
+        vm.prank(owner);
+        quest.queue();
+
+        vm.expectRevert(abi.encodeWithSelector(NotEnded.selector));
```
Quazia, line 268: Might be good to standardize the error messages between this and `Quest.t.sol` should we capture this in a separate issue? Out of scope for this PR.

Quazia (approved):
Looks good - I would prefer we don't use the magic numbers in a few places and can we capture the fuzzed tests as a follow on.

PR #210, opened 2023-09-21
add QuestFactory.t.sol

waynehoover:
- Adds QuestFactory.t.sol

### DIFF
--- a/test/QuestFactory.t.sol
+++ b/test/QuestFactory.t.sol
@@ -99,7 +99,7 @@ contract TestQuestFactory is Test, Errors, Events, TestUtils {
         vm.stopPrank();
     }
 
-    function test_RevertIf_reate1155QuestAndQueue_MsgValueLessThanQuestNFTFee() public {
+    function test_RevertIf_create1155QuestAndQueue_MsgValueLessThanQuestNFTFee() public {
         vm.startPrank(questCreator);
 
         sampleERC1155.mintSingle(questCreator, 1, TOTAL_PARTICIPANTS);


test/QuestFactory.t.sol
```diff
@@ -0,0 +1,537 @@
...
+            "questId",
+            "actionSpec",
+            0
+        );
+    }
+
+    /*//////////////////////////////////////////////////////////////
+                                CLAIM
+    //////////////////////////////////////////////////////////////*/
+    function test_claim_with_claim1155Rewards_with_referrer() public{
```
Quazia, line 232: `QuestTypeNotSupported` this revert isn't caught

Quazia (changes requested):
Mostly looks good I think a couple small cases arn't tested. I think `processMintFee` and `returnChange` should be fuzz tested by way of one of the functions that call them. Overall looks good though. Let me know if you think those two would be better served in a follow-on, main reason I think a fuzz is worth it here is the exposure of accounting in those functions.

test/QuestFactory.t.sol
```diff
@@ -99,6 +99,24 @@ contract TestQuestFactory is Test, Errors, Events, TestUtils {
         vm.stopPrank();
     }
 
+    function test_RevertIf_reate1155QuestAndQueue_MsgValueLessThanQuestNFTFee() public {
```
Quazia, line 102: typo in function name "reate" vs "create"

Quazia (approved):
LGTM - small typo but I like the fuzz claim logic   ☺️

PR #213, opened 2023-09-26
QuestFactory move the mint fee to init

waynehoover:
I noticed this while working o QuestFactory.t.sol: https://github.com/rabbitholegg/quest-protocol/pull/210/files#diff-4754184314c26ca979d48f490abdcb0de1066e4ee18ef1ca0b27eec9bc4f284cR65

There isn't really a reason not to do this in the initializer now, because it's always going to be set.


Quazia (approved):
LGTM

PR #221, opened 2023-10-16
add chainid and timestamp to quest address seed

waynehoover:
This will make quest addresses unique, this helps us move forward with using the [quest address as the id](https://github.com/rabbitholegg/quest-protocol/pull/219), and some infra changes in the indexer.

Why not use CREATE over CREATE2 as CREATE will always give a unique address? Good question, because with create there is a small possibility of a reorg attack to create the same quest address after a re-org. This is mostly for chains like Polygon where reorgs happen more frequently than one would expect.


Quazia (approved):
LGTM

PR #222, opened 2023-10-17
add new claim function and QuestClaimedData event

waynehoover:
- adds `function claim(bytes calldata signature_, bytes calldata data_)`
- adds  `QuestClaimedData` event
- claimer can now be passed to this new `claim` function, it is no longer assumes `msg.sender` is the claimer.


.solhint.json
```diff
@@ -1,7 +1,8 @@
 {
-  "extends": "solhint:recommended", 
+  "extends": "solhint:recommended",
   "rules": {
-    "compiler-version": ["error",">=0.8.0"], 
-    "avoid-low-level-calls": "off"
+    "compiler-version": ["error",">=0.8.0"],
+    "avoid-low-level-calls": "off",
+    "quotes": ["off","double"]
```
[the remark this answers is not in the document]
Quazia: Another option would be to ignore that file, or ignore that rule for that file - I'm neutral

Quazia (approved):
LGTM 

Is the idea to eventually remove
```
function claim(string memory questId_, bytes32 hash_, bytes memory signature_, address ref_) external payable
```
If that's used as an entry-point we don't get the extra information

test/QuestFactory.t.sol
```diff
@@ -395,6 +395,52 @@ contract TestQuestFactory is Test, Errors, Events, TestUtils {
...
+        vm.warp(START_TIME + 1);
+
+        string memory json = string(abi.encodePacked('{"anyting": "we want", "foo": "bar"}'));
+
+        bytes memory data = abi.encode(participant, address(0), "questId2", json);
+        bytes32 msgHash = keccak256(data);
+        bytes memory signature = signHash(msgHash, claimSignerPrivateKey);
+
+        vm.startPrank(participant);
+        vm.recordLogs();
```
Quazia, line 425: This is a cool testing pattern for the multiple logs I like this.

test/QuestFactory.t.sol
```diff
@@ -395,6 +395,52 @@ contract TestQuestFactory is Test, Errors, Events, TestUtils {
...
+        // erc20 reward
+        assertEq(sampleERC20.balanceOf(participant), REWARD_AMOUNT, "particpiant erc20 balance");
+
+        // claim fee rewards
+        assertEq(questCreator.balance - questCreatorBeforeBalance, MINT_FEE / 3, "questCreator mint fee");
+        assertEq(protocolFeeRecipient.balance, (MINT_FEE / 3) * 2, "protocolFeeRecipient mint fee");
+
+        // assert QuestClaimedData event
+        Vm.Log[] memory entries = vm.getRecordedLogs();
+        assertEq(entries.length, 5);
```
Quazia, line 437: Maybe add reason strings to these asserts. Could be hard to tell which section is failing otherwise.

Quazia (approved):
LGTM just some small optional nits and a typo

PR #225, opened 2023-10-31
Add test to confirm json output, and remove unneeded event

waynehoover:
- Removes `ClaimedSingle` event, as it's not used and redundant.
- Adds a test to confirm the JSON string emitted is valid JSON.


Quazia (approved):
LGTM - I don't love the `returnedJson.parse();` but also don't have a better solution and I think having it is better than not having it.

PR #226, opened 2023-11-03
Fix typo and upgrade everywhere

waynehoover:


Quazia (approved):
LGTM - that's a lot of LOC for an extra `f` 😆

PR #228, opened 2023-11-13
Gas Efficient Claim Functions

waynehoover:
- Removes legacy `claim` functions, and deprecates `claim` function
- Adds new `claimOptimized` function.

Deploy this after BE goes out, and make sure to set `OPTIMIZED_CLAIM_FUNCTION_CUTOFF_TIMESTAMP` env there.


contracts/Quest1155.sol
```diff
@@ -121,6 +126,50 @@ contract Quest1155 is ERC1155Holder, ReentrancyGuardUpgradeable, PausableUpgrade
...
+    function recoverSigner(bytes32 hash_, bytes memory signature_) public view returns (address) {
+        return ECDSA.recover(ECDSA.toEthSignedMessageHash(hash_), signature_);
+    }
+
+    /// @dev universal dynamic claim function
+    /// @param signature_ The signature of the data
+    /// @param data_ The data to decode for the claim
+    function claim(bytes calldata signature_, bytes calldata data_) external payable
+        // nonReentrant need to be sure we don't need this
+        whenNotEnded
```
Quazia, line 141: Is there a way to setup the quest creators withdraw function to remove the need for the whenNotEnded for the claim?

contracts/Quest1155.sol
```diff
@@ -121,6 +126,50 @@ contract Quest1155 is ERC1155Holder, ReentrancyGuardUpgradeable, PausableUpgrade
...
+        onlyStarted
```
Quazia, line 142: Same here - theoretically if we only sign for quests that have started can we remove this check too?

contracts/Quest1155.sol
Quazia: I'm aligned with your thinking - not worth it.

contracts/Quest.sol
```diff
@@ -254,7 +245,11 @@ contract Quest is ReentrancyGuardUpgradeable, PausableUpgradeable, Ownable, IQue
     /*//////////////////////////////////////////////////////////////
                              INTERNAL VIEW
     //////////////////////////////////////////////////////////////*/
-    function _receiptRedeemers() internal view returns (uint256) {
+    function _redeemedTokens() internal view returns (uint256) {
         return questFactoryContract.getNumberMinted(questId);
     }
+
+    function _claimFee() internal view returns (uint256) {
+        return questFactoryContract.mintFee();
```
Quazia, line 253: We could potentially cache this value or pass it through in the `claimFromFactory` call to prevent the callback. Obviously caching it is functionally different.

contracts/QuestFactory.sol
```diff
@@ -328,33 +317,41 @@ contract QuestFactory is Initializable, LegacyStorage, OwnableRoles, IQuestFacto
...
+
+        emit QuestClaimedData(claimer_, msg.sender, jsonData_);
+        if (quest.questType.eq("erc1155")) {
+            emit Quest1155Claimed(claimer_, msg.sender, questId_, rewardToken_, tokenId_);
+        } else {
+            emit QuestClaimed(claimer_, msg.sender, questId_, rewardToken_, tokenId_);
+        }
+        if(ref_ != address(0)){
+            emit QuestClaimedReferred(claimer_, msg.sender, questId_, rewardToken_, tokenId_, ref_, 3333, mintFee);
+            emit MintFeePaid(questId_, address(0), 0, address(0), 0, ref_, mintFee / 3); // check to be sure needed
```
Quazia, line 353: What's this comment here?

test/Quest.t.sol
```diff
@@ -254,95 +256,42 @@ contract TestQuest is Test, TestUtils, Errors, Events {
...
-            SampleERC20(rewardTokenAddress).balanceOf(address(questFactoryMock)),
-            questBalanceMinusFees,
-            "quest should have 0 tokens"
+            protocolFeeRecipient.balance,
+            CLAIM_FEE * TOTAL_PARTICIPANTS * 1 / 3,
+            "owner should have received remaining ETH"
         );
-    }
 
-    function test_fuzz_withdrawRemainingTokens(
```
Quazia, line 277: Is there a reason we're getting rid of the fuzz testing?

Quazia (approved):
LGTM - iit'd be good to have some understanding around the fuzz tests that are removed. Is that related to the changes to the math where the protocol basically gets whatever is remaining? Small potential optimization around the `mintFee` but we can look into that in a follow on when more of the event flow gets simplified.

PR #229, opened 2023-11-15
Various gas saving measures for the claim function

waynehoover:
Some small refactoring that save about 3k gas.

- reduce gas by 
  - not doing the same checks in the factory as in the singleClaim function on each quest contract
  - removing the NFT fee from NFT quests as this is not used 
- [ ] remove `numberMinted` from questFactory and use `redeemedTokens` on quest instead.

- renames some errors to be consistent on each quest contract


contracts/QuestFactory.sol
```diff
@@ -605,12 +602,9 @@ contract QuestFactory is Initializable, LegacyStorage, OwnableRoles, IQuestFacto
     {
         Quest storage currentQuest = quests[claimData_.questId];
         IQuestOwnable questContract_ = IQuestOwnable(currentQuest.questAddress);
-        if (!questContract_.queued()) revert QuestNotQueued();
-        if (block.timestamp < questContract_.startTime()) revert QuestNotStarted();
-        if (block.timestamp > questContract_.endTime()) revert QuestEnded();
```
jonathandiep, line 610: I think removing this line will allow people to be able to claim past the quest end time

PR #244, opened 2023-12-25
Add compressed claim function to Quest contracts

waynehoover:
This will bring our calldata down from 212 bytes to 116 (with a referrer) a saving of 45.28%.

- Adds a highly calldata optimized `claim` function to the Quest and Quest1155 contracts via the abstract contract `QuestClaimable`.
- Overloads the create functions to include the `txHashChainId` in quest creation.
- Includes some small refactors
- As usual, this is backwards compatible.


contracts/libraries/QuestClaimable.sol
```diff
@@ -0,0 +1,32 @@
...
+pragma solidity ^0.8.19;
+
+import { IQuestFactory } from "../interfaces/IQuestFactory.sol";
+
+abstract contract QuestClaimable {
+    function getQuestFactoryContract() public view virtual returns (IQuestFactory);
+
+    function getQuestId() public view virtual returns (string memory);
+
+    function claim() external payable {
```
Quazia, line 11: How does this get handled when settled to L1 and how does the gas used work in this call?

Quazia (approved):
LGTM - that new `claim` without any calldata is crazy can you comment that some more? I don't fully understand how the L1 will account for that/how the gas used will be calculated.
Small nit `Depricated` is spelled wrong a few places.

PR #246, opened 2023-12-27
Add Soulbound20 Contract

waynehoover:
This adds a soulbound ERC20 contract to be used as points. 

Assumptions:
- Many addresses should be able to mint at a time.
- Only the owner can change who these address are that can mint
- Only the owner can enable transfers
- The owner is the QuestFactory (can be transferred)


Quazia (approved):
LGTM - we can remove the basic Ownable import. Don't need the minter event now. Also some comments only relevant to upgradable contracts. Should we target a feature branch for all the points work or just target main and wait to deploy?

PR #250, opened 2023-12-30
Add tx orgin check to all external claim functions

waynehoover:
This adds a tx.origin check to all external claim functions, with exceptions for being called from the quest factory or being called from a quest contract.

### DIFF
--- a/contracts/QuestFactory.sol
+++ b/contracts/QuestFactory.sol
@@ -358,8 +358,6 @@ contract QuestFactory is Initializable, LegacyStorage, OwnableRoles, IQuestFacto
     /// @param data_ The claim data in abi encoded bytes
     /// @param signature_ The signature of the claim data
     function claimOptimized(bytes calldata signature_, bytes calldata data_) external payable {
-        if(tx.origin != msg.sender) revert txOriginMismatch();
-
         (
             address claimer_,
             address ref_,


contracts/QuestFactory.sol
```diff
@@ -346,6 +344,9 @@ contract QuestFactory is Initializable, LegacyStorage, OwnableRoles, IQuestFacto
 
         string memory questIdString_ = bytes16ToUUID(questid_);
         Quest storage quest_ = quests[questIdString_];
+
+        if(tx.origin != msg.sender && tx.origin != quest_.questAddress) revert txOriginMismatch();
```
Quazia, line 348: What's the case where the origin is actually the quest?

contracts/QuestFactory.sol
```diff
@@ -369,6 +368,9 @@ contract QuestFactory is Initializable, LegacyStorage, OwnableRoles, IQuestFacto
             (address, address, string, string)
         );
         Quest storage quest = quests[questId_];
+
+        if(tx.origin != msg.sender && tx.origin != address(this) && tx.origin != quest.questAddress) revert txOriginMismatch();
```
Quazia, line 372: For the `address(this)` case, we should be able to split this into an internal helper `_claimOptimized` and an external to get around needing to do this extra check. Also same question for what case is the `tx.origin` actually the `questAddress`.

Quazia (approved):
This LGTM - I'm seeing now that it's the `QuestClaimable` with that super tightened up `claim` with no params that's requiring the `tx.origin` I think we could potentially drop some of the extra guards by having internal functions that are called through different external functions that have different/more lax checks (require `tx.origin == quest.questAddress`) But I'm not sure that's an improvement. Also I think some of the `tx.origin != address(this)` statements might not be needed, I think we really only need that statement in the `QuestClaimable` `claim()` function as it stands right now assuming we don't use the split internal helpers.

PR #254, opened 2024-01-05
msg.sender changes, not tx.origin

waynehoover:


Quazia (approved):
LGTM only concern is with the unit testing. Are we testing that it does throw `txOriginMismatch` when it's not the same origin/sender?

PR #256, opened 2024-01-16
using env file test_signer for all tests

gnomadic:


README.md
```diff
@@ -122,6 +122,14 @@ forge install
 forge build
 ```
 
+### Create .env file
+
+>  this uses a compromised private key from hardhat's test node, do not use this key for anything other than automated testing!
+
+```bash
+cp .env.example .env
+sed -i '' "s/TEST_CLAIM_SIGNER_PRIVATE_KEY=/TEST_CLAIM_SIGNER_PRIVATE_KEY=0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80/" .env
```
gnomadic, line 131: holler if we should remove this line and let people figure it out on their own!

test/QuestFactory.t.sol
```diff
@@ -247,18 +247,20 @@ contract TestQuestFactory is Test, Errors, Events, TestUtils {
...
-        bytes32 txHash = 0x57498a77018f78c02a0e2f0d0e4a8aab048b6e249ff936d230b7db7ca48782e1;
         uint32 txHashChainId = 1;
-        bytes16 questId = 0x88e08cb195e64832845fa92ec8f2034a;
-        string memory questIdString = "88e08cb1-95e6-4832-845f-a92ec8f2034a";
-        string memory actionType = "other";
-        string memory questName = "Advanced Trading with dYdX";
-        bytes32 r = 0x12a1078fd9cf9bbed1fa8f00b9abd75baa6c07073706479ca25ec34f4043b327;
-        bytes32 vs = 0x5aca8bfe397d45aa673c07d2ce845cb4419eadc0cbe8977de337799a37b2e1a3;
+        bytes32 txHash = hex'57498a77018f78c02a0e2f0d0e4a8aab048b6e249ff936d230b7db7ca48782e1';
+        bytes16 questId = hex'88e08cb195e64832845fa92ec8f2034a';
```
jonathandiep, line 252: Is there any benefit for using `hex'...'`? Or is it just preference?

Quazia (approved):
LGTM - I'm assuming doing the hard coded in-line was an effort to get around the stack-too-deep issue. I'd prefer using variables and either relying on a new context or internal functions to get around the stack depth issues. You should be able to just add `{}` to give yourself a new stack context, or as a last resort a testing struct would reduce the number of variables needed. Or at the least I would try and have the `questIdString` be a variable since that is used in multiple places. Another option would be global constants or setup block variables.

test/QuestFactory.t.sol
gnomadic: I'm under the impression just preference for readability and a compilation error if we copy/paste incorrectly.

I was trying to be consistent with the existing code and saw the hex literal used pretty commonly

gnomadic:
Is there a reason to use different chain IDs, quest IDs, quest names, and action types across different individual tests?

If they're just used for the signature verification on chain, then we're probably good to have one small test to validate our signature validation with different permutations and then every other test can use common globals for those values...?

If that sounds right to you I can get that in this PR and clean this up more

Quazia:
@gnomadic That sounds like a great pattern. I think a base-case `struct` with the repeated params would be the most readable but however you want to do it sounds like an improvement.

gnomadic:
@Quazia I ended up adding a struct definition to the utils, and creating an instance globally - then every test can use it easily without worrying about breaking stuff.  I also wanna refactor out [these events](https://github.com/rabbitholegg/quest-protocol/blob/main/test/helpers/Events.sol) and instead use the productions ones so that we don't miss any event changes in our tests.... but that's for another day

I ended up not writing a test for the signing code, as I dug a bit deeper and it looks like it's just solidity and LibZip.. so I think we have to assume those two functions are well tested already

PR #264, opened 2024-02-09
BoostPassVotes

jonathandiep:


.env.example
```diff
@@ -3,6 +3,9 @@ MAINNET_PRIVATE_KEY=
 TEST_CLAIM_SIGNER_PRIVATE_KEY=
 MAINNET_CLAIM_SIGNER_PRIVATE_KEY=
 
+ALCHEMY_ARBITRUM_RPC=
+ALCHEMY_SEPOLIA_RPC=
```
Quazia, line 7: Gotta add these to the github settings so the test pass - I can do this if needed just lmk

contracts/BoostPass.sol
```diff
@@ -19,6 +20,7 @@ contract BoostPass is Initializable, Ownable, ERC721 {
     uint256 public mintFee;
     address public treasuryAddress;
     string private baseURI;
+    address private boostPassVotesAddress;
```
Quazia, line 23: Looks good should preserve storage order

contracts/BoostPassVotes.sol
```diff
@@ -0,0 +1,209 @@
...
+    }
+
+    function _transferVotingUnits(address from, address to, uint256 amount) internal virtual {
+        if (from == address(0)) {
+            _push(_totalCheckpoints, _add, SafeCastLib.toUint224(amount));
+        }
+        if (to == address(0)) {
+            _push(_totalCheckpoints, _subtract, SafeCastLib.toUint224(amount));
+        }
+        // we don't need to move delegate votes upon mint
```
Quazia, line 144: I'd remove this

contracts/BoostPassVotes.sol
```diff
@@ -0,0 +1,209 @@
...
+        // _moveDelegateVotes(delegates(from), delegates(to), amount);
+    }
+
+    function _moveDelegateVotes(address from, address to, uint256 amount) private {
+        if (from != to && amount > 0) {
+            if (from != address(0)) {
```
Quazia, line 150: If we use a separate internal function for the case where address(0) is to/from we could avoid needing to do these checks. Something about this logic flow feels off to me. If we keep it as it is I'd add a comment/callout for when and why address(0) is. Something like "When initially minting we don't want to change the delegated votes of address(0)" idk.

contracts/BoostPassVotes.sol
Quazia: Also, if we don't delegate voting rights on mint can you just talk me through when a delegation occurs to/from address(0)?

test/BoostPassVotes.t.sol
```diff
@@ -0,0 +1,260 @@
...
+        // Execute proposal
+        uint256 deadline = myGovernor.proposalDeadline(proposalId);
+        vm.warp(deadline + 1);
+        uint256 successfulProposalId = myGovernor.execute(targets, values, calldatas, keccak256(abi.encodePacked(description)));
+
+        assertEq(successfulProposalId, proposalId);
+        assertEq(dai.balanceOf(owner), 1 ether);
+    }
+
+    function test_delegate() public {
```
Quazia, line 200: I'd like to see some multi-step delegation tests, what it looks like when someone "undelegates" and some testing around what happens when address(0) is passed into delegate. Ideally some fuzz testing where we iterate through a stack of delegators but that may be too heavy here. Regardless undelegated and multi-step delegation would be good to test.

Quazia (changes requested):
This is looking _so good_ thank you 🙏 

I think we want some more testing around delegation, and maybe to massage some of the logic there depending on results of testing.

I'd also use copilot or some tool assist to generate the natspec for any public/external function at minimum.

PR #269, opened 2024-03-19
remove tx.origin checks

ccashwell:
Changes:
- Removed `tx.origin` validation
- Removed origin mismatch revert test
  - Incidentally, the claim that was being passed in for that test fails anyway, so that was replaced as well.


test/QuestFactory.t.sol
```diff
@@ -306,8 +305,7 @@ contract TestQuestFactory is Test, Errors, Events, TestUtils {
     }
 
     function test_claimCompressed_revert_txOriginMismatch() public{
```
Quazia, line 307: Should we just remove the whole test?

Quazia (approved):
LGTM - I'd remove the whole test but not mission critical I reckon.

test/QuestFactory.t.sol
ccashwell: Yeah, I always feel iffy about deleting tests that assert specific behavior like this. Open to removing it.

Quazia: Well we arn't asserting that behavior anymore, right?

PR #270, opened 2024-03-21
Feat(QuestFactory): Remove Token Whitelist

Quazia:


contracts/QuestFactory.sol
```diff
@@ -89,10 +89,9 @@ contract QuestFactory is Initializable, LegacyStorage, OwnableRoles, IQuestFacto
                                MODIFIERS
     //////////////////////////////////////////////////////////////*/
 
-    modifier checkQuest(string memory questId_, address rewardTokenAddress_) {
+    modifier checkQuest(string memory questId_) {
         Quest storage currentQuest = quests[questId_];
         if (currentQuest.questAddress != address(0)) revert QuestIdUsed();
-        if (!rewardAllowlist[rewardTokenAddress_]) revert RewardNotAllowed();
```
jonathandiep, line 95: Should we also set a comment on line 44 that `mapping(address => bool) public rewardAllowlist;` is not used

jonathandiep (approved):
lgtm but left a comment!

Also, wondering if we should remove `setRewardAllowlistAddress`

ccashwell:
lgtm 👌

PR #274, opened 2024-04-04
Referral Rewards

jonathandiep:


contracts/Quest.sol
```diff
@@ -135,11 +143,15 @@ contract Quest is ReentrancyGuardUpgradeable, PausableUpgradeable, Ownable, IQue
...
         _transferRewards(account_, totalRedeemableRewards);
+
     }
 
     function claimFromFactory(address claimer_, address ref_) external payable whenNotEnded onlyQuestFactory {
         _transferRewards(claimer_, rewardAmountInWei);
-        if (ref_ != address(0)) ref_.safeTransferETH(_claimFee() / 3);
+        if (ref_ != address(0)) {
+            ref_.safeTransferETH(_claimFee() / 3);
+            _updateReferralTokenAmount(ref_);
```
Quazia, line 153: This feels wrong to transfer some out and to account for the other internally

contracts/Quest.sol
```diff
@@ -155,7 +167,7 @@ contract Quest is ReentrancyGuardUpgradeable, PausableUpgradeable, Ownable, IQue
         protocolFeeRecipient.safeTransferETH(protocolPayout);
 
         // transfer reward tokens
-        uint256 protocolFeeForRecipient = this.protocolFee() / 2;
+        uint256 protocolFeeForRecipient = (this.protocolFee() / 2) - referralClaimTotal;
```
Quazia, line 170: Nice approach - clean

contracts/Quest.sol
```diff
@@ -164,6 +176,17 @@ contract Quest is ReentrancyGuardUpgradeable, PausableUpgradeable, Ownable, IQue
         questFactoryContract.withdrawCallback(questId, protocolFeeRecipient, protocolPayout, address(owner()), ownerPayout);
     }
 
+    function claimReferralFees(address referrer) external onlyWithdrawAfterEnd {
+        if (referrerHasClaimed[referrer] == true) revert AlreadyWithdrawn();
+
+        uint256 referrerClaimAmount = referralClaimAmounts[referrer];
+        if (referrerClaimAmount > 0) {
```
Quazia, line 183: I would either do this check as a revert to prevent the tx from even getting pushed or just remove it altogether.

contracts/Quest.sol
```diff
@@ -218,6 +249,14 @@ contract Quest is ReentrancyGuardUpgradeable, PausableUpgradeable, Ownable, IQue
         rewardToken.safeTransfer(sender_, amount_);
     }
 
+    /// @notice Internal function to update the referral reward amount
+    /// @param referrer_ The address of the referrer
+    function _updateReferralTokenAmount(address referrer_) internal {
+        uint256 referralAmount = this.referralRewardAmount();
```
Quazia, line 255: Let's push this into an internal function that gets called here and in the external function:
```suggestion
        uint256 referralAmount = _referralRewardAmount();
```

Then we do...
```
 function referralRewardAmount() external view returns (uint256) {
        return _refferalRewardMount();
    }
    function _referralRewardAmount() internal view returns (uint256) {
        return (referralRewardFee * rewardAmountInWei) / 10_000;
    }
```
    
This prevents us from "calling out" an external call triggered with `this.` will behave differently. Less of an issue when there's no calldata but still better form.

contracts/Quest1155.sol
```diff
@@ -180,6 +181,10 @@ contract Quest1155 is ERC1155Holder, ReentrancyGuardUpgradeable, PausableUpgrade
         return questId;
     }
 
+    function referralRewardAmount() external pure returns (uint256) {
```
Quazia, line 184: I would confirm we wanna do 0 for these. Makes sense but let's confirm with product just so there are no surprises.

contracts/QuestFactory.sol
```diff
@@ -166,7 +172,8 @@ contract QuestFactory is Initializable, LegacyStorage, OwnableRoles, IQuestFacto
                 actionType_,
                 questName_,
                 "erc20",
-                projectName_
+                projectName_,
+                referralRewardFee_
```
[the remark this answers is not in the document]
Quazia: Can't we consistently derive it from stuff we're already throwing?

contracts/QuestFactory.sol
```diff
@@ -419,7 +428,29 @@ contract QuestFactory is Initializable, LegacyStorage, OwnableRoles, IQuestFacto
             emit QuestClaimed(claimer_, quest.questAddress, questId_, rewardToken_, rewardAmountOrTokenId);
         }
         if(ref_ != address(0)){
-            emit QuestClaimedReferred(claimer_, quest.questAddress, questId_, rewardToken_, rewardAmountOrTokenId, ref_, 3333, mintFee);
+            if (IQuestOwnable(quest.questAddress).startTime() > referralRewardTimestamp) {
+                emit QuestClaimReferred(
```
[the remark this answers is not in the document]
Quazia: SGTM -  might be easier to just emit empty data though, keeps our data shaped more consistently. I don't have a strong opinion depends a lot more on ease-of-consumption.

contracts/QuestFactory.sol
```diff
@@ -484,6 +515,10 @@ contract QuestFactory is Initializable, LegacyStorage, OwnableRoles, IQuestFacto
         defaultMintFeeRecipient = mintFeeRecipient_;
     }
 
+    function setReferralRewardTimestamp(uint256 timestamp_) external onlyOwner {
```
Quazia, line 518: Need to remember to call this as part of the deployment. Another option is to actually just set this in the initialize call.

test/Quest.t.sol
```diff
@@ -302,6 +308,92 @@ contract TestQuest is Test, TestUtils, Errors, Events {
         quest.withdrawRemainingTokens();
     }
 
+    /*//////////////////////////////////////////////////////////////
+                        CLAIM REFERRAL FEES
+    //////////////////////////////////////////////////////////////*/
+
+    function test_claimReferralFees() public {
```
Quazia, line 315: Can we fuzz this test just to check edges? Preference for fuzzed tests whenever there's accounting in play.

Quazia (changes requested):
LGTM - only things blocking are the fuzzed test and not doing external calls using `this.`

contracts/QuestFactory.sol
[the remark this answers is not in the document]
jonathandiep: Hmm I think we can keep the same event. We may need to update our db model in order to store the value though

contracts/QuestFactory.sol
jonathandiep: It's a part of the `initialize` call already, but I think we are going to need to call this function on chains where we upgrade

contracts/Quest1155.sol
jonathandiep: Will confirm, but don't think we can give a percentage of 1155s lol

contracts/Quest.sol
jonathandiep: Awesome thanks for this

test/Quest.t.sol
jonathandiep: I updated this to be `testFuzz_claimReferralFees()` and based off the results when running `forge test`, I don't think it's actually running the fuzz tests

`test_fuzz_claimReferralFees()` doesn't work either -- lets sync up on this

ccashwell: The name isn't important, you have to pass in a fuzzable arg:

```
function test_someFuzzableFunction(uint256 fuzzableArg) {
  // ...
  contract.someMutativeFunctionWithArguments(fuzzableArg);
  // ...
}
```

But it doesn't really make sense here because the only input is an address which only enables a binary logic branch: either the address has fees to claim or it does not. The purpose of fuzz testing is mainly to catch cases off the happy path that unexpectedly break the contract's invariants. 

Generally, you'd want to either mock out something the function depends on to return a fuzzed value (rather than enforcing the correct execution path by performing the "normal" operations inline as pre-assertion setup), or accept arguments which can be manipulated by the fuzz test. The only thing that seems fuzzable to me in this specific function is the time, since you're warping around and expecting certain behavior depending on the time.

Quazia: Less concerned with the address more concerned with the accounting around the referral payouts since this is the only test function that's covering those accounting paths.

Quazia (approved):
LGTM - we can add fuzzing for `referralRewardFee_` as a follow-on, doesn't need to be blocking.

contracts/Quest.sol
```diff
@@ -170,7 +170,7 @@ contract Quest is ReentrancyGuardUpgradeable, PausableUpgradeable, Ownable, IQue
         uint256 protocolFeeForRecipient = (this.protocolFee() / 2) - referralClaimTotal;
         rewardToken.safeTransfer(protocolFeeRecipient, protocolFeeForRecipient);
 
-        uint256 remainingBalanceForOwner = rewardToken.balanceOf(address(this));
+        uint256 remainingBalanceForOwner = rewardToken.balanceOf(address(this)) - referralClaimTotal;
```
Quazia, line 173: This doesn't account for the fact that some claims might have gone through already.

Gonna need to update the value in the claim too. I would maintain to variables:
`referralClaimTotal` and `totalReferralsClaimed` or something like this.

Here we want:
`        uint256 remainingBalanceForOwner = rewardToken.balanceOf(address(this)) - (referralClaimTotal - totalReferralsClaimed) ;
`

Quazia (changes requested):

PR #277, opened 2024-05-13
feat(quests): implement cancel operation for quests [BOOST-3960]

ccashwell:
This adds the ability to `cancel` a Quest at any point before the quest's natural end time. 

In order to facilitate cancellation without introducing new complexity, I chose to reuse the existing `pause` functionality (which, incidentally, wasn't actually being used anywhere). When the quest owner calls `cancel`, the quest is transitioned into `paused` state, which has the effect of preventing any future changes to the end time as the cancel function is gated by the `whenNotPaused` modifier. If the quest has not yet started (i.e. `startTime > block.timestamp`), the quest is cancelled effective immediately. If it has already started, the `endTime` is changed to be the current block timestamp plus 15 minutes to allow for any in-flight actions to resolve. The time delay was added to address the potential for malicious quest creators to abuse cancellations to deprive unsuspecting users of earned incentives.

### DIFF
--- a/contracts/Quest1155.sol
+++ b/contracts/Quest1155.sol
@@ -105,10 +105,12 @@ contract Quest1155 is ERC1155Holder, ReentrancyGuardUpgradeable, PausableUpgrade
     /*//////////////////////////////////////////////////////////////
                             EXTERNAL UPDATE
     //////////////////////////////////////////////////////////////*/
-    /// @notice Pauses the Quest
-    /// @dev Only the owner of the Quest can call this function. Also requires that the Quest has started (not by date, but by calling the start function)
-    function pause() external onlyOwner onlyStarted {
+
+    /// @notice Cancels the Quest by setting the end time to 15 minutes from the current time and pausing the Quest. If the Quest has not yet started, it will end immediately.
+    /// @dev Only the owner of the Quest can call this function.
+    function cancel() external onlyOwner whenNotPaused whenNotEnded {
         _pause();
+        endTime = startTime > block.timestamp ? block.timestamp : block.timestamp + 15 minutes;
     }
 
     /// @notice Queues the quest by marking it ready to start at the contract level. Marking a quest as queued does not mean that it is live. It also requires that the start time has passed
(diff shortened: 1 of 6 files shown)


jonathandiep (approved):
lgtm!

ccashwell:
@jonathandiep @Quazia can I get a second pass on this? implemented changes to improve indexability and in such. BOOST-3972 + BOOST-3960

contracts/Quest1155.sol
```diff
@@ -105,10 +105,12 @@ contract Quest1155 is ERC1155Holder, ReentrancyGuardUpgradeable, PausableUpgrade
...
     //////////////////////////////////////////////////////////////*/
-    /// @notice Pauses the Quest
-    /// @dev Only the owner of the Quest can call this function. Also requires that the Quest has started (not by date, but by calling the start function)
-    function pause() external onlyOwner onlyStarted {
+
+    /// @notice Cancels the Quest by setting the end time to 15 minutes from the current time and pausing the Quest. If the Quest has not yet started, it will end immediately.
+    /// @dev Only the owner of the Quest can call this function.
+    function cancel() external onlyQuestFactory whenNotPaused whenNotEnded {
         _pause();
+        endTime = startTime > block.timestamp ? block.timestamp : block.timestamp + 15 minutes;
```
benjaminykim, line 113: why 15 minutes? This means that user can't withdraw until 15 minutes after cancellation correct?

ccashwell: Yep, the delay is intended to protect users by allowing a 15-minute window during which any in-flight completions can be indexed and paid out.

Quazia (approved):
Looks good - I like the added making sure they can cancel right away if it hasn't started. Does what it says it does on the can 👍

jonathandiep (approved):

contracts/QuestFactory.sol
```diff
@@ -329,6 +329,14 @@ contract QuestFactory is Initializable, LegacyStorage, OwnableRoles, IQuestFacto
         );
     }
 
+    function cancelQuest(string calldata questId_) external {
+        Quest storage _questData = quests[questId_];
+        if (_questData.questCreator != msg.sender) revert Unauthorized();
+        IQuestOwnable quest = IQuestOwnable(_questData.questAddress);
+        quest.cancel();
+        emit QuestCancelled(_questData.questAddress, questId_, quest.endTime());
```
jonathandiep, line 337: @ccashwell Can you double check? I think this event is emitting the old `endTime`

jonathandiep: Some context: https://sepolia.etherscan.io/tx/0x0c9044b4deef12007d85ea74059a642da4e7a15c2c8f1a1fc8dada820b2cbb82#eventlog

The value of `1716526800` is not 15 mins from block time (at time of this comment)

jonathandiep: Actually I think I know why this is! This Quest contract was created before the latest changes

On a new Quest deployment, the logic is working as expected 👍

PR #284, opened 2024-06-18
feat: modify referral fees

jonathandiep:


contracts/Quest.sol
```diff
@@ -80,13 +78,13 @@ contract Quest is ReentrancyGuardUpgradeable, PausableUpgradeable, Ownable, IQue
...
         questFee = questFee_;
         protocolFeeRecipient = protocolFeeRecipient_;
-        referralRewardFee = referralRewardFee_;
 
         // Setup default state
         questFactoryContract = IQuestFactory(payable(msg.sender));
         queued = true;
         referralClaimTotal = 0;
         totalReferralsFeesClaimed = 0;
+        referralRewardFee = 250; // 2.5%
```
Quazia, line 87: Why hardcode vs pass it in?

test/Quest.t.sol
```diff
@@ -322,15 +300,15 @@ contract TestQuest is Test, TestUtils, Errors, Events {
                         CLAIM REFERRAL FEES
     //////////////////////////////////////////////////////////////*/
 
-    function test_fuzz_claimReferralFees(uint96 timestamp, uint256 participants, uint256 referralRewardFee) public {
+    function test_fuzz_claimReferralFees(uint96 timestamp, uint256 participants) public {
         timestamp = uint96(bound(timestamp, START_TIME+10, END_TIME));
         participants = bound(participants, 1, TOTAL_PARTICIPANTS);
-        referralRewardFee = bound(referralRewardFee, 1, REFERRAL_REWARD_FEE);
+        // referralRewardFee = bound(referralRewardFee, 1, REFERRAL_REWARD_FEE);
```
Quazia, line 306: delete

test/Quest.t.sol
```diff
@@ -432,15 +409,15 @@ contract TestQuest is Test, TestUtils, Errors, Events {
         );
     }
 
-    function test_fuzz_claimReferralFees_withdrawAfterClaim(uint96 timestamp, uint256 participants, uint256 referralRewardFee) public {
+    function test_fuzz_claimReferralFees_withdrawAfterClaim(uint96 timestamp, uint256 participants) public {
         timestamp = uint96(bound(timestamp, START_TIME+10, END_TIME));
         participants = bound(participants, 1, TOTAL_PARTICIPANTS);
-        referralRewardFee = bound(referralRewardFee, 1, REFERRAL_REWARD_FEE);
+        // referralRewardFee = bound(referralRewardFee, 1, REFERRAL_REWARD_FEE);
```
Quazia, line 415: ``

test/helpers/TestUtils.sol
```diff
@@ -8,24 +8,34 @@ contract TestUtils is Test {
...
-    function calculateTotalFees(
+    function calculateTotalProtocolFees(
         uint256 totalParticipants,
         uint256 rewardAmount,
         uint16 questFee
     ) internal pure returns (uint256) {
         return (totalParticipants * rewardAmount * questFee) / 10_000;
     }
 
+    function calculateTotalAffiliateFees(
```
Quazia, line 31: Where is this used and what is it used for?

Quazia (changes requested):
The code looks good and the switching of internal/external look right but the use of affiliate vs referral is sort of confusing so can you just state out the desired target for the changes and when it's affiliate vs referral?

Quazia (approved):
LGTM -  confirmed the withdrawal logic too just to be safe. Everything checks out.

contracts/Quest.sol
jonathandiep: We are going to hardcode it, but we could pass it in (and move hardcode to the QuestFactory instead)

Quazia (approved):

PR #289, opened 2024-07-09
Update QuestBudget Fee Calculation

jonathandiep:


contracts/QuestFactory.sol
```diff
@@ -85,6 +86,7 @@ contract QuestFactory is Initializable, LegacyStorage, OwnableRoles, IQuestFacto
         referralFee = referralFee_;
         mintFee = mintFee_;
         referralRewardTimestamp = block.timestamp;
+        referralRewardFee = 250; // in BIPS
```
jonathandiep, line 89: The `referralFee` on line 86 refers to the referral fees on the claim fee (which is currently set to 0 ETH)

Quazia (approved):
Looks good :shipit:

PR #290, opened 2024-07-09
feat: add claimCompressedRefBatch function

veganbeef:
This PR adds a batch version of `claimCompresedRef` to allow for multiple claims on behalf of multiple claimers at once.

Individual claim failures do not cause the transaction to revert but do emit a new `BatchClaimFailed` event.

This is to support batch claims for Drakula ([linear ticket](https://linear.app/rh-app/issue/BOOST-4228/add-batch-version-of-claimcompressedref-to-boost-contracts)).

### DIFF
--- a/contracts/QuestFactory.sol
+++ b/contracts/QuestFactory.sol
@@ -405,7 +405,6 @@ contract QuestFactory is Initializable, LegacyStorage, OwnableRoles, IQuestFacto
             }
         }
 
-        // Refund any excess ETH
         uint256 excess = msg.value - totalFees;
         if (excess > 0) {
             payable(msg.sender).transfer(excess);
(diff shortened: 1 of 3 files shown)


contracts/QuestFactory.sol
```diff
@@ -388,6 +388,29 @@ contract QuestFactory is Initializable, LegacyStorage, OwnableRoles, IQuestFacto
         _claimCompressed(compressedData_, claimer);
     }
 
+    /// @dev Claim rewards for multiple quests on behalf of multiple claimers
+    function claimCompressedRefBatch(BatchClaimData[] calldata batchClaimDataArray) external payable {
+        uint256 totalFees = 0;
+        for (uint256 i = 0; i < batchClaimDataArray.length; i++) {
+            totalFees += batchClaimDataArray[i].fee;
+        }
```
jonathandiep, line 396: I think we can calculate totalFees by doing `batchClaimDataArray.length * mintFee`

jonathandiep: Doing this would also allow us to get rid of `fee` from the `BatchClaimData` struct

